PR #1, opened 2020-02-05
[user] Add basic user service with GET endpoint

jaylees14:
Adds a basic GET handler to a user service and build script.

To run:

```
~/g/s/g/T/spec-golang ❯❯❯ sh build.sh user
```

Testing:
```
~/g/s/g/T/spec-golang ❯❯❯ curl localhost:80/user 
"Hello, World"
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+docker build -t $1 -f $1/Dockerfile $1
+docker run -p 80:80 $1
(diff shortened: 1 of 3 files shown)


c-harding (approved):
Can’t say I speak Go, but it all looks sensible. The idea of functions returning “error” seems a little odd, but what can you do when you can’t throw?

willjvsmith (approved):
LGTM 🚀

PR #2, opened 2020-02-06
[user] Add Postgres Docker container and fetch values via GET request

jaylees14:
- Configure Postgres container
- Fetch values from `GET` request based on URL

Testing (will add unit tests later on, seemed less important for now):
```
~/g/s/g/T/spec-golang ❯❯❯ sh build.sh
```
```
~/g/s/g/T/spec-golang ❯❯❯ curl localhost/user/1
{"ID":1,"Name":"Andrew"}
```
```
~/g/s/g/T/spec-golang ❯❯❯ curl localhost/user/2                                                                                           
Something went wrong: User for ID 2 does not exist
```

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -2,19 +2,39 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"log"
 	"net/http"
+	"strconv"
 
+	"github.com/TempleEight/spec-golang/user/dao"
 	"github.com/gorilla/mux"
 )
 
 func main() {
 	r := mux.NewRouter()
-	r.HandleFunc("/user", userGetHandler).Methods(http.MethodGet)
+	r.HandleFunc("/user/{id}", userGetHandler).Methods(http.MethodGet)
 
 	log.Fatal(http.ListenAndServe(":80", r))
 }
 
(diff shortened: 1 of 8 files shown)


user/user.go
```diff
@@ -2,19 +2,39 @@ package main
...
+
+	userID, err := strconv.ParseInt(userIDStr, 10, 64)
+	if err != nil {
+		http.Error(w, "Invalid user ID provided", http.StatusBadRequest)
+		return
+	}
+
+	user, err := dao.GetUser(userID)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Something went wrong: %s", err.Error()), http.StatusInternalServerError)
```
c-harding: User doesn’t exist sounds like a 4xx (404?), not a 500

c-harding (approved):

user/user.go
jaylees14: Yeah good point, I guess there's actually multiple errors that could happen here, so might be worth defining those types and switching on them

Lewiky (approved):
err err err.err err err

willjvsmith (approved):
Perfect 👌

PR #4, opened 2020-02-09
[matches] Add basic matches service

jaylees14:
Add a service for "matches" - this will combine a match between two "users".

Currently this is exposed via port 81, however I'll work on an API Gateway next that unifies access to the services

```
~/g/s/g/T/spec-golang ❯❯❯ docker-compose up --build
```
```
~/g/s/g/T/spec-golang ❯❯❯ curl localhost:81/matches
"Hello, World"

~/g/s/g/T/spec-golang ❯❯❯ curl localhost:80/user/1
{"ID":1,"Name":"Andrew"}
```

NB: You may need to tidy up the `Postgres` volumes to test from a fresh DB:

```
~/g/s/g/T/spec-golang ❯❯❯ docker volume prune -f
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,5 +17,23 @@ services:
     networks: 
       - user-network 
 
+  matches:
+    build: ./matches
+    ports:
+      - "81:80" 
+    networks: 
+      - matches-network 
+    
+  matches-db:
+    image: postgres:12.1 
+    environment:
+      - PGUSER=postgres
+    volumes:
+      - ./matches-db/init.sql:/docker-entrypoint-initdb.d/init.sql
+    networks: 
+      - matches-network 
+  
+
 networks:
   user-network:
+  matches-network:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -17,5 +17,23 @@ services:
...
+      - PGUSER=postgres
+    volumes:
+      - ./matches-db/init.sql:/docker-entrypoint-initdb.d/init.sql
+    networks: 
+      - matches-network 
+  
+
 networks:
   user-network:
+  matches-network:
```
c-harding, line 39: Why are these empty keys rather than a list?

jaylees14: That just defines the network (https://docs.docker.com/compose/networking/) - Don't think we need to do anything custom with them, so they're just empty for now

willjvsmith (approved):
LGTM 🚀

Lewiky (approved):

PR #6, opened 2020-02-10
[user] Add update & delete endpoints

jaylees14:
- Add update and delete endpoints to users
- Update endpoint is a `PUT` so requires all parameters to be provided

Setup:
```
$ curl -X GET localhost/user/2
{"error":"user not found with ID 2"}

$ curl -X POST localhost/user -d '{"name":"Jay"}'
{"ID":2,"Name":"Jay"}

$ curl -X GET localhost/user/2
{"ID":2,"Name":"Jay"}
```

Updating:
```
$ curl -X PUT localhost/user/2 -d '{"name":"Lewis"}'
{}

$ curl -X GET localhost/user/2
{"ID":2,"Name":"Lewis"}
```

Deleting:
```
$ curl -X DELETE localhost/user/2
{}

$ curl -X GET localhost/user/2
{"error":"user not found with ID 2"}
```

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"strconv"
 
 	"github.com/TempleEight/spec-golang/user/dao"
 	valid "github.com/asaskevich/govalidator"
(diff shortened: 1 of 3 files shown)


user/user.go
```diff
@@ -83,3 +78,60 @@ func userCreateHandler(w http.ResponseWriter, r *http.Request) {
...
+	err = json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		errMsg := CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	_, err = valid.ValidateStruct(req)
+	if err != nil {
+		errMsg := CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
```
Lewiky, line 99: Can this whole block be abstracted out? Seems like we do it a lot - maybe for another PR though

Lewiky (approved):
LGTM

user/user.go
c-harding: Is this happening this PR or next?

jaylees14: Probably the next

c-harding (approved):

PR #7, opened 2020-02-10
[api-gateway] Add basic API gateway

jaylees14:
Add an API Gateway, such that requests can be directed to a single endpoint.
This PR makes a couple of changes:
- Changes the exposed port of matches service to 81
- Adds an additional network for the API Gateway to share with each service - you can visualise the network infrastructure as:

```
             +-----------------------------------------------+ kong-network
             |                                               |
             |                                               |
             |                                               |
             |                                               |
             |                                               |
             |                    kong                       |
             |                                               |
             |                                               |
             |                                               |
        +----|---------------------+   +---------------------|-----+
        |    |                     |   |                     |     |
        |    |                     |   |                     |     |
        |    |    user-svc         |   |      matches-svc    |     |
        |    |                     |   |                     |     |
        |    |                     |   |                     |     |
        |    +-----------------------------------------------+     |
        |                          |   |                           |
        |                          |   |                           |
        |                          |   |                           |
        |                          |   |                           |
        |    user-postgres         |   |      matches-postgres     |
        |                          |   |                           |
        |                          |   |                           |
        |                          |   |                           |
        +--------------------------+   +---------------------------+
        user-network                                matches-network
```
- This stops the two Postgres instances from interacting → Not sure if this is the best way to do it, will probably change in the future if we actually deploy the services to something other than localhost

To test:
```
$ docker-compose up --build
$ sh kong/configure-kong.sh
```
```
$ curl localhost:8000/api/user/1
{"ID":1,"Name":"Andrew"}
```
```
$ curl localhost:8000/api/matches
"Hello, World"
```

The services are also individually accessible via `localhost:80` and `localhost:81`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ services:
       - "80:80" 
     networks: 
       - user-network 
+      - kong-network
 
   user-db:
     image: postgres:12.1 
@@ -20,9 +21,10 @@ services:
   matches:
     build: ./matches
     ports:
-      - "81:80" 
+      - "81:81" 
     networks: 
       - matches-network 
+      - kong-network
     
   matches-db:
     image: postgres:12.1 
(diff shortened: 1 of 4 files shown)


c-harding (approved):

Lewiky (approved):
LGTM

willjvsmith (approved):
LGTM 🚀 

Obvs not final, but good until we expand this example service

PR #8, opened 2020-02-11
[matches] Add CRUD endpoints

Lewiky:
Full implementation of all the CRUD endpoints for the matches service - nearly directly copied from what the user service already does

Features a lot of repeated code, could we abstract this out? Discussion still ongoing

Also the first time I've ever written Go, so would appreciate feedback.

Test plan:

```
> curl -X GET localhost:8000/api/matches/1
{"IDs":[1]}
```

```
> curl -X GET localhost:8000/api/match/1
{"ID":1,"UserOne":1,"UserTwo":2,"MatchedOn":<TIME>}
```

```
> curl -X POST localhost:8000/api/match -d '{"UserOne":3, "UserTwo":4}'
{"ID":2,"UserOne":3,"UserTwo":4,"MatchedOn":<TIME>"}
```

```
> curl -X DELETE localhost:8000/api/match/1
{}
```

```
> curl -X PATCH localhost:8000/api/match -d '{"id":2, "UserOne":5, "UserTwo":6}'
{}
```


matches-db/init.sql
```diff
@@ -1 +1,8 @@
-CREATE TABLE Matches (userOne int, userTwo int);
+CREATE TABLE Matches (
+    id SERIAL PRIMARY KEY,
+    userOne int,
+    userTwo int,
+    matchedOn timestamp
+  );
+
+INSERT INTO Matches (userOne, userTwo, matchedOn) VALUES (1, 2, NOW());
```
c-harding, line 8: Why is population happening in the init script?

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
+package dao
+
+import (
+	// pq acts as the driver for SQL requests
```
c-harding: Comment seems to have got lost

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+
+// MatchGetResponse contains the information stored about a given match
+type MatchGetResponse struct {
+	ID        int
+	userOne   int
+	userTwo   int
+	matchedOn string
+}
+
+// MatchListResponse contains the information stored about all matches
```
c-harding, line 21: Is there any point in opening every comment with the function/type name?

matches/dao/errors.go
```diff
@@ -0,0 +1,10 @@
+package dao
+
+import "fmt"
+
+// ErrMatchNotFound is returned when a user for the provided ID was not found
+type ErrMatchNotFound int64
+
+func (e1 ErrMatchNotFound) Error() string {
```
jaylees14: Would just call that variable e

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+		default:
+			return nil, err
+		}
+	}
+
+	return &match, nil
+}
+
+// ListMatch lists all the matches which feature a user
+func ListMatch(id1 int64) (*MatchListResponse, error) {
```
jaylees14: Would call that id probably

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+	ID      int
+	userOne int
+	userTwo int
+}
+
+func executeQueryWithResponses(query string, args ...interface{}) (*sql.Rows, error) {
+	db, err := sql.Open("postgres", connStr)
+	if err != nil {
+		return nil, err
+	}
```
c-harding, line 51: Thoughts on this, to make the error handling more clearly bound to the source?

```go
if db, err := sql.Open("postgres", connStr); err != nil {
	return nil, err
}
```

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+
+// ListMatch lists all the matches which feature a user
+func ListMatch(id1 int64) (*MatchListResponse, error) {
+	rows, err := executeQueryWithResponses("SELECT id FROM Matches WHERE userOne = $1 OR userTwo = $1", id1)
+	if err != nil {
+		return nil, err
+	}
+
+	var matches MatchListResponse
+	matches.IDs = make([]int, 0)
```
jaylees14, line 110: Can we make a list of the correct size using the rows object? Not sure if it returns the number of rows found

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	return &matches, nil
+}
+
+func CreateMatch(request MatchCreateRequest) (*MatchCreateResponse, error) {
+	row, err := executeQueryWithRowResponse("INSERT INTO Matches (userOne, userTwo, matchedOn) VALUES ($1, $2, NOW()) RETURNING *", request.ID1, request.ID1)
```
jaylees14: Think that second argument should be .id2

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+		default:
+			return nil, err
+		}
+	}
+
+	return &match, nil
+}
+
+func UpdateMatch(request MatchUpdateRequest) error {
+	rowsAffected, err := executeQuery("UPDATE Matches SET userOne = $1, userTwo = $2 WHERE id = $3", request.userTwo, request.userOne, request.ID)
```
jaylees14: Think the arguments are the wrong way around, should be userOne then userTwo

matches/dao/match-dao.go
c-harding: It feels like NOW() should be calculated in Go for simplicity

matches/matches.go
```diff
@@ -2,19 +2,161 @@ package main
...
 
+	"github.com/TempleEight/spec-golang/matches/dao"
+	valid "github.com/asaskevich/govalidator"
 	"github.com/gorilla/mux"
 )
 
 func main() {
 	r := mux.NewRouter()
-	r.HandleFunc("/matches", matchGetHandler).Methods(http.MethodGet)
+	r.HandleFunc("/matches/{id1}", matchListHandler).Methods(http.MethodGet)
```
jaylees14: Would avoid using 1 here, just stick with id

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+	var match MatchCreateResponse
+	err = row.Scan(&match.ID, &match.userOne, &match.userTwo, &match.matchedOn)
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+			return nil, errors.New("Could not create match")
+		default:
+			return nil, err
+		}
+	}
```
c-harding, line 149: Are these eight lines the most gooey they can be? Seems extremely verbose, or at least space-inefficient

tinder.temple
```diff
@@ -0,0 +1,36 @@
+Tinder: project {
```
jaylees14: Probably wouldn’t commit this, since wasn’t generated with it. Could put it in the readme?

matches/dao/match-dao.go
jaylees14: This should probably go above the pq import

matches/matches.go
```diff
@@ -2,19 +2,161 @@ package main
...
+func matchUpdateHandler(w http.ResponseWriter, r *http.Request) {
+	var req dao.MatchUpdateRequest
+	err := json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		errMsg := CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	_, err = valid.ValidateStruct(req)
```
c-harding, line 143: Hmmm, why does `ValidateStruct` return a boolean if you never use it?

c-harding (approved):
Just some comments, looks fine though

jaylees14 (commented):
Nice one - couple of things: I’d personally break this down into separate pull requests in the future. Try and keep the content minimal and focussed on one goal only. 

Other than that, looks good - just a couple of minor things, and I’d stick with one of “userOne” or “user1”. Personally prefer the former, but up to you

matches/dao/match-dao.go
jaylees14: Required by GoFmt

Lewiky: Think this is required by the go formatter

matches/matches.go
```diff
@@ -2,19 +2,161 @@ package main
...
+	"github.com/TempleEight/spec-golang/matches/dao"
+	valid "github.com/asaskevich/govalidator"
 	"github.com/gorilla/mux"
 )
 
 func main() {
 	r := mux.NewRouter()
-	r.HandleFunc("/matches", matchGetHandler).Methods(http.MethodGet)
+	r.HandleFunc("/matches/{id1}", matchListHandler).Methods(http.MethodGet)
+	r.HandleFunc("/match/{id}", matchGetHandler).Methods(http.MethodGet)
```
jaylees14, line 18: Think these require a new route to be added to the Kong configuration since the url differs from the one previously defined

c-harding:
Keyboard with dedicated Go key

matches/matches.go
Lewiky: Think it's just the err we need

matches/dao/match-dao.go
Lewiky: https://stackoverflow.com/questions/23415612/insert-datetime-using-now-with-go

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+
+// MatchListResponse contains the information stored about all matches
+type MatchListResponse struct {
+	IDs []int
+}
+
+// MatchCreateRequest contains the information required to create a new match
+type MatchCreateRequest struct {
+	ID1 string `valid:"type(string),required,stringlength(1|255)"`
+	ID2 string `valid:"type(string),required,stringlength(1|255)"`
```
jaylees14: `idOne` to stick with same naming as others

matches/dao/match-dao.go
willjvsmith: Not a particularly strong feeling, but prefer as is

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+type MatchUpdateRequest struct {
+	ID      int
+	userOne int
+	userTwo int
+}
+
+func executeQueryWithResponses(query string, args ...interface{}) (*sql.Rows, error) {
+	db, err := sql.Open("postgres", connStr)
```
willjvsmith, line 48: 🐌

I guess we'll change this in the tidy-up

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return result.RowsAffected()
+}
+
+// GetMatch returns the information about a match stored for a given ID
+func GetMatch(id int64) (*MatchGetResponse, error) {
+	row, err := executeQueryWithRowResponse("SELECT * FROM Matches WHERE id = $1", id)
```
willjvsmith, line 83: funky dollars

(alternative is question marks)

user-db/init.sql
```diff
@@ -4,3 +4,5 @@ CREATE TABLE Users (
 );
 
 INSERT INTO Users (name) VALUES ('Andrew');
+
+INSERT INTO Users (name) VALUES ('Schien');
```
willjvsmith, line 8: 😠

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+
+// MatchCreateResponse contains the information stored about the newly created match
+type MatchCreateResponse struct {
+	ID        int
+	userOne   int
+	userTwo   int
+	matchedOn string
+}
+
+type MatchUpdateRequest struct {
```
jaylees14, line 41: This needs a `godoc` string

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	return &matches, nil
+}
+
+func CreateMatch(request MatchCreateRequest) (*MatchCreateResponse, error) {
```
jaylees14, line 134: Needs `godoc`

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+			return nil, errors.New("Could not create match")
+		default:
+			return nil, err
+		}
+	}
+
+	return &match, nil
+}
+
+func UpdateMatch(request MatchUpdateRequest) error {
```
jaylees14, line 155: Needs `godoc`

matches/dao/match-dao.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return err
+	} else if rowsAffected == 0 {
+		return ErrMatchNotFound(request.ID)
+	}
+
+	return nil
+}
+
+func DeleteMatch(id int64) error {
```
jaylees14, line 167: Needs `godoc`

jaylees14 (commented):
LGTM 👍

matches/matches.go
```diff
@@ -2,19 +2,161 @@ package main
...
+			http.Error(w, errMsg, http.StatusInternalServerError)
+		}
+		return
+	}
+
+	json.NewEncoder(w).Encode(match)
+}
+
+func matchListHandler(w http.ResponseWriter, r *http.Request) {
+	userOneStr := mux.Vars(r)["id"]
```
willjvsmith: Perhaps rename to just `userStr` since can be userOne or userTwo in the match, and similarly throughout the rest of this func

willjvsmith (commented):
Nice

willjvsmith (approved):
LGTM 🚀

willjvsmith:
Added test plan to PR description.

Also to note:
* The create endpoint returns the row created, however update does not. This matches the current user service implementation, but we'll address this in the forthcoming tidy-up.
* We probs want to check for duplicate matches when creating, though I guess this is more a business logic issue.

matches/matches.go
```diff
@@ -2,19 +2,161 @@ package main
...
 )
 
 func main() {
 	r := mux.NewRouter()
-	r.HandleFunc("/matches", matchGetHandler).Methods(http.MethodGet)
+	r.HandleFunc("/matches/{id}", matchListHandler).Methods(http.MethodGet)
+	r.HandleFunc("/match/{id}", matchGetHandler).Methods(http.MethodGet)
+	r.HandleFunc("/match", matchCreateHandler).Methods(http.MethodPost)
+	r.HandleFunc("/match/{id}", matchDeleteHandler).Methods(http.MethodDelete)
+	r.HandleFunc("/match", matchUpdateHandler).Methods(http.MethodPatch)
```
jaylees14, line 21: This is a `PUT`, not a `PATCH`: [see here](https://rapidapi.com/blog/put-vs-patch/)

matches/dao/match-dao.go
```diff
@@ -0,0 +1,176 @@
...
+
+// MatchListResponse contains the information stored about all matches
+type MatchListResponse struct {
+	IDs []int
+}
+
+// MatchCreateRequest contains the information required to create a new match
+type MatchCreateRequest struct {
+	UserOne int
+	UserTwo int
```
jaylees14, line 29: This allows me to provide no request body and still create a match - since it assumes the default values of 0 for each integer. Should this be allowed? 

```
❯❯❯ curl -X POST localhost:8000/api/match -d '{}'
{"ID":3,"UserOne":0,"UserTwo":0,"MatchedOn":"2020-02-12T14:36:21.62474Z"}
```

jaylees14 (commented):
Almost there - this should definitely have been broken down into separate PRs. It's far too large to review in one go.

matches/dao/match-dao.go
willjvsmith: I asked the same thing in the Slack chat, I think not, sounds like you agree

jaylees14: Assuming that `SERIAL` is used as the key type in the `User` table, it can never be 0 (https://www.postgresqltutorial.com/postgresql-serial/) - so we could validate it has to be >= 1?

matches/matches.go
willjvsmith: That link supports that the current implementation should be a `PATCH`, a partial update. as the `matchedOn` field is not changed. Also it doesn't create a new entry if it doesn't exist.

However leaving the `matchedOn` field unchanged (and not giving the option to update it) seems wrong. Allowing this to be updated, and creating a new entry if the ID doesn't exist, would make it a `PUT` as you say.

I agree that this PR is far too large, there are lots of small decisions being made here and things being overlooked.

matches/dao/match-dao.go
willjvsmith: I think if the field isn't passed explicitly in the body then it shouldn't change.

willjvsmith: > I think if the field isn't passed explicitly in the body then it shouldn't change.

There must be a nice way of doing this, I'll look into it in the tidy-up.

matches/matches.go
jaylees14: I think this one is quite ambiguous since a request could never update the `matchedOn` field, which means the update request body matches that of the `POST` request. To an end user of this service this the would suggest entire resource is updated, hence I suggested `PUT`. 

`PATCH` would also imply that the user could only provide certain fields to the body and only these fields would be updated: this isn't the case at all, since we update all fields using the request contents.

With regards to creating the entity if it doesn't exist, the RFC has some interesting wording around that implying it's a choice rather than a requirement:

> If the Request-URI does not point to an existing resource, and that URI is capable of being defined as a new resource by the requesting user agent, the origin server **can** create the resource with that URI.

We can leave it as is now, but I think I'd actually support removing the update functionality all together.

jaylees14 (approved):

PR #9, opened 2020-02-13
[tidy] Optimise DAO and add configuration file handling for User service

willjvsmith:
* Optimises User service DAO to no longer creates a new database connection per request
* Adds configuration file handling, JSON format
  * Command-line argument `-config` for specifying configuration file path
* Adds Kong script to build script
* Removes `Andrew` and `Schien` inserts in `init.sql`

Sozzles for this large PR.

Test plan:

```
> ./build.sh
```
```
> curl -X GET localhost:8000/api/user/1
{"error":"user not found with ID 1"}
> curl -X POST localhost:8000/api/user -d '{"name":"Will"}'
{"ID":1,"Name":"Will"}
> curl -X GET localhost:8000/api/user/1
{"ID":1,"Name":"Will"}
> curl -X PUT localhost:8000/api/user/1 -d '{"name":"Charlie"}'
{}
> curl -X GET localhost:8000/api/user/1
{"ID":1,"Name":"Charlie"}
> curl -X DELETE localhost:8000/api/user/1
{}
> curl -X GET localhost:8000/api/user/1
{"error":"user not found with ID 1"}
```


c-harding:
> Sozzles for this **large PR**

Do I pick up the tiniest hint of passive aggression?

user/utils/utils.go
```diff
@@ -1,11 +1,28 @@
...
 
 import (
 	"encoding/json"
 	"errors"
+	"os"
 	"strconv"
 )
 
+// GetConfig returns a configuration object from decoding the given configuration file
+func GetConfig(filePath string) (Config, error) {
```
jaylees14: I’d personally have this return `(*Config, error)` so you can avoid returning an invalid config if there’s an error

user/config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "user": "postgres",
+  "dbName": "postgres",
+  "host": "user-db",
+  "sslMode": "disable"
+}
```
c-harding: ` ⃠⤶`
Sort your editor out

user/dao/user-dao.go
```diff
@@ -35,34 +37,40 @@ type UserUpdateRequest struct {
...
 		return 0, err
 	}
-	defer db.Close()
 
-	result, err := db.Exec(query, args...)
+	return result.RowsAffected()
+}
+
+// Initialise opens the database connection
+func (dao *DAO) Initialise(configFilePath string) error {
```
jaylees14: I’d be tempted to pass in the config here and do the file IO in main. Keeps all the side effects in one place and should be easier to test if we want to

jaylees14 (commented):
Looks good, just a few comments

jaylees14:
Oh and another thought - do you ever close the database connection?

Lewiky (approved):
l a r g e b o I

user/dao/user-dao.go
```diff
@@ -53,13 +53,9 @@ func executeQuery(db *sql.DB, query string, args ...interface{}) (int64, error)
...
 // Initialise opens the database connection
-func (dao *DAO) Initialise(configFilePath string) error {
-	config, err := utils.GetConfig(configFilePath)
-	if err != nil {
-		return err
-	}
-
+func (dao *DAO) Initialise(config *utils.Config) error {
 	connStr := fmt.Sprintf("user=%s dbname=%s host=%s sslmode=%s", config.User, config.DBName, config.Host, config.SSLMode)
+	var err error
```
c-harding, line 58: This mix of `:=` and `var`+`=` is a bit horrible :′(

c-harding (approved):

willjvsmith:
> Oh and another thought - do you ever close the database connection?

[Quick search suggests we don't have to](https://www.reddit.com/r/golang/comments/4f5mjf/when_should_i_close_database_connection/), though we probably want to add a mechanism for graceful shutdown. Can add it in there.

user/dao/user-dao.go
willjvsmith: ikr 😢

c-harding:
> > Oh and another thought - do you ever close the database connection?
> 
> [Quick search suggests we don't have to](https://www.reddit.com/r/golang/comments/4f5mjf/when_should_i_close_database_connection/), though we probably want to add a mechanism for graceful shutdown. Can add it in there.

This is what I was talking about when I said Go was accidentally well designed. This database library seems almost impossible to go make a mistake with

jaylees14 (approved):
LGTM, I think graceful shutdown is definitely worth looking at - we’ll probably want to stop additional requests but ensure we finish any that are pending. That’s for another PR though.

PR #11, opened 2020-02-17
[tidy] Misc fixes to User and Match services

willjvsmith:
* Assigns a fixed version to Kong docker image
* Reorders functions/structs to follow CRUD
* Adds an ignore flag, `valid:"-"`, to Match structs
  * This means govalidator's `validateStruct` can be called on the struct without causing an error
  * Though we don't need to call `validateStruct` on them anyway, so could take out (but would mean more effort code generating later"
* Forces calls to CREATE and UPDATE to pass all fields
  * Makes CREATE and UPDATE request struct fields pointers, so it can be discerned if they were omitted (otherwise they take zero-value)
* Renames `Users` table to `User_Temple`
  * Avoids pluralising problem, e.g. `User` -> `Users`, `Match` -> `Matches`
  * Unfortunately `User` is reserved in Postgres, `Temple` suffix handles this
    * Decision made to maintain a map of reserved names when generating
* Removes User and Match CREATE endpoints returning the row, returns `{}` to match UPDATE
* Change Match UPDATE to be a HTTP PUT rather than PATCH
* Rename `ExtractUserIDFromRequest` in User service to `ExtractIDFromRequest`, to match the same function in Match service (less code to generate)

The diff here is not the best because of the reordering of functions/structs to follow CRUD. If too hard to follow LMK and I'll put the reordering in a separate PR. I've tried to detail everything above to mitigate.

#### Test plan

```
> ./build.sh
> sh kong/configure-kong.sh
```

User service:
```
> curl -X GET localhost:8000/api/user/1
{"error":"user not found with ID 1"}
> curl -X POST localhost:8000/api/user -d '{"name":"Will"}'
{}
> curl -X GET localhost:8000/api/user/1
{"ID":1,"Name":"Will"}
> curl -X PUT localhost:8000/api/user/1 -d '{"name":"Charlie"}'
{}
> curl -X GET localhost:8000/api/user/1
{"ID":1,"Name":"Charlie"}
> curl -X DELETE localhost:8000/api/user/1
{}
> curl -X GET localhost:8000/api/user/1
{"error":"user not found with ID 1"}
```

Match service:
```
> curl -X GET localhost:8000/api/match/1
{"error":"match not found with ID 1"}
> curl -X POST localhost:8000/api/match -d '{"UserOne":1, "UserTwo":2}'
{}
> curl -X GET localhost:8000/api/match/1
{"ID":1,"UserOne":1,"UserTwo":2,"MatchedOn":<TIME>}
> curl -X PUT localhost:8000/api/match/1 -d '{"UserOne":3, "UserTwo":4}'
{}
> curl -X GET localhost:8000/api/match/1
{"ID":1,"UserOne":3,"UserTwo":0,"MatchedOn":<TIME>}
> curl -X DELETE localhost:8000/api/match/1
{}
> curl -X GET localhost:8000/api/match/1
{"error":"match not found with ID 1"}
```

Missing fields:
```
> curl -X POST localhost:8000/api/match -d '{"UserOne":5}'
{"error":"Missing request parameter(s)"}
> curl -X POST localhost:8000/api/match -d '{"UserOne":5, "UserTwo":6}'
{}
> curl -X PUT localhost:8000/api/match/1 -d '{"UserOne":7}'
{"error":"Missing request parameter(s)"}
```

### DIFF
--- a/match/dao/match-dao.go
+++ b/match/dao/match-dao.go
@@ -2,7 +2,6 @@ package dao
 
 import (
 	"database/sql"
-	"errors"
 	"fmt"
 
 	"github.com/TempleEight/spec-golang/match/utils"
@@ -15,6 +14,12 @@ type DAO struct {
 	DB *sql.DB
 }
 
+// MatchCreateRequest contains the information required to create a new match
+type MatchCreateRequest struct {
+	UserOne int `valid:"-"`
+	UserTwo int `valid:"-"`
+}
+
 // MatchGetResponse contains the information stored about a given match
 type MatchGetResponse struct {
 	ID        int
@@ -23,30 +28,15 @@ type MatchGetResponse struct {
 	MatchedOn string
 }
 
+// MatchUpdateRequest contains the information required to update a match
+type MatchUpdateRequest struct {
+	UserOne int `valid:"-"`
+	UserTwo int `valid:"-"`
+}
+
 // MatchListResponse contains the information stored about all matches
 type MatchListResponse struct {
 	IDs []int
-}
-
-// MatchCreateRequest contains the information required to create a new match
-type MatchCreateRequest struct {
-	UserOne int
-	UserTwo int
-}
-
-// MatchCreateResponse contains the information stored about the newly created match
-type MatchCreateResponse struct {
-	ID        int
-	UserOne   int
-	UserTwo   int
-	MatchedOn string
-}
-
-// MatchUpdateRequest contains the information required to update a match
-type MatchUpdateRequest struct {
-	ID      int
-	UserOne int
-	UserTwo int
 }
 
 // Executes the query, returning the rows
@@ -81,6 +71,12 @@ func (dao *DAO) Initialise(config *utils.Config) error {
 	return nil
 }
 
+// CreateMatch inserts a new match into the database given two user IDs
+func (dao *DAO) CreateMatch(request MatchCreateRequest) error {
+	_, err := executeQuery(dao.DB, "INSERT INTO Match (userOne, userTwo, matchedOn) VALUES ($1, $2, NOW()) RETURNING *", request.UserOne, request.UserTwo)
+	return err
+}
+
 // GetMatch returns the information about a match stored for a given ID
 func (dao *DAO) GetMatch(id int64) (*MatchGetResponse, error) {
 	row, err := executeQueryWithRowResponse(dao.DB, "SELECT * FROM Match WHERE id = $1", id)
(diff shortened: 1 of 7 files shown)


match/match.go
```diff
@@ -52,6 +52,36 @@ func jsonMiddleware(next http.Handler) http.Handler {
...
 
+func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
+	var req matchDAO.MatchCreateRequest
+	err := json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+	if req.UserOne == nil || req.UserTwo == nil {
```
willjvsmith, line 63: Rather than doing this, I was thinking of writing a function that takes any struct and checks if any of the pointers are `nil`. This would make generation easier, but perhaps at the cost of readability.

I looked into this and it seems achievable [using reflection](https://stackoverflow.com/questions/49466522/how-to-check-if-all-fields-of-a-struct-are-nil). I thought I would get others' thoughts before going for it though. Lewis and I suspect Charlie will be all for, but Jay against.

match/dao/match-dao.go
```diff
@@ -102,6 +98,30 @@ func (dao *DAO) GetMatch(id int64) (*MatchGetResponse, error) {
 	return &match, nil
 }
 
+// UpdateMatch updates an already existing match to two user IDs
+func (dao *DAO) UpdateMatch(matchID int64, request MatchUpdateRequest) error {
+	rowsAffected, err := executeQuery(dao.DB, "UPDATE Match SET userOne = $1, userTwo = $2 WHERE id = $3", request.UserOne, request.UserTwo, matchID)
```
jaylees14: Probably update the `matchedOn` time here too

match/match.go
```diff
@@ -34,10 +34,10 @@ func main() {
 	}
 
 	r := mux.NewRouter()
-	r.HandleFunc("/match/{id}", matchGetHandler).Methods(http.MethodGet)
 	r.HandleFunc("/match", matchCreateHandler).Methods(http.MethodPost)
+	r.HandleFunc("/match/{id}", matchGetHandler).Methods(http.MethodGet)
 	r.HandleFunc("/match/{id}", matchDeleteHandler).Methods(http.MethodDelete)
-	r.HandleFunc("/match", matchUpdateHandler).Methods(http.MethodPatch)
+	r.HandleFunc("/match/{id}", matchUpdateHandler).Methods(http.MethodPut)
 	r.HandleFunc("/matches/{id}", matchListHandler).Methods(http.MethodGet)
```
jaylees14, line 41: Should also update this to `/match/all`

match/match.go
jaylees14: I think this is fine

jaylees14: https://www.calhoun.io/how-to-determine-if-a-json-key-has-been-set-to-null-or-not-provided/ Could be a solution, although might again be less readable: although, would avoid using reflection ;)

jaylees14 (approved):
LGTM

match/match.go
willjvsmith: Will do in new PR.

c-harding (approved):
looks good

PR #12, opened 2020-02-18
[tidy] Fix all matches endpoint

willjvsmith:
* Removes `matches` route, replace with `match/all`
* Makes `match/all` return all matches, including all fields for each one
* Identifiers are named to avoid pluralism problem, `match` -> `matches`

While it doesn't make sense for this sample service, we decided to omit the endpoint for fetching all matches for a user as it leads to high complexity in our DSL and subsequently in the parser. For now we're sticking with just generating CRUD, GET all, and GET all enumerable by one field endpoints. Shout out to the Oxford comma.

#### Test plan

```
> ./build.sh
> sh kong/configure-kong.sh
```

```
> curl -X GET localhost:8000/api/match/all
{"MatchList":[]}
> curl -X POST localhost:8000/api/match -d '{"UserOne":1, "UserTwo":2}'
{}
> curl -X GET localhost:8000/api/match/all
{"MatchList":[{"ID":1,"UserOne":1,"UserTwo":2,"MatchedOn":<TIME>}]}
```

### DIFF
--- a/match/dao/match-dao.go
+++ b/match/dao/match-dao.go
@@ -122,7 +122,7 @@ func (dao *DAO) DeleteMatch(id int64) error {
 	return nil
 }
 
-// ListMatch lists all matches
+// ListMatch returns a list containing every match
 func (dao *DAO) ListMatch() (*MatchListResponse, error) {
 	rows, err := executeQueryWithResponses(dao.DB, "SELECT * FROM Match")
 	if err != nil {
--- a/match/dao/match-dao.go
+++ b/match/dao/match-dao.go
@@ -36,7 +36,7 @@ type MatchUpdateRequest struct {
 
 // MatchListResponse contains the information stored about all matches
 type MatchListResponse struct {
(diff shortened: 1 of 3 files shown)


Lewiky (approved):

match/dao/match-dao.go
```diff
@@ -122,33 +122,28 @@ func (dao *DAO) DeleteMatch(id int64) error {
 	return nil
 }
 
-// ListMatch lists all the matches which feature a user
-func (dao *DAO) ListMatch(id int64) (*MatchListResponse, error) {
-	rows, err := executeQueryWithResponses(dao.DB, "SELECT id FROM Match WHERE userOne = $1 OR userTwo = $1", id)
+// ListMatch lists all matches
```
c-harding: "every match"

c-harding (approved):
Looking good

jaylees14 (approved):

PR #13, opened 2020-02-19
[Go-Gen] Implement Cross-Service Communication

Lewiky:
This PR performs basic Cross-Service communication between match service and user service. Namely to check that both users exist before trying to make a match between them.

Provides a new package for handling the communication, it reads from the config file the hostnames of all of the services it cares about.

While implementing this PR I also wrote some code to act as logging middleware, it was decided that this should be taken out and should be done properly at another time, but I saved what I wrote in a gist [here](https://gist.github.com/Lewiky/0ca7c63823c0658f35715639f1245fd4), more information can be found at [this SO post](https://stackoverflow.com/questions/43021058/golang-read-request-body). 

## Test Plan

After setting up the docker system in the usual way:

``` bash
$> curl -X POST http://localhost:8000/api/match -d '{"UserOne": 1, "UserTwo": 2}'

{"error":"Unknown Users: 1 | 2"}

$> curl -X POST http://localhost:8000/api/user -d '{"Name": "Lewis"}'

{}

$> curl -X POST http://localhost:8000/api/user -d '{"Name": "Jay"}'

{}

$> curl -X POST http://localhost:8000/api/match -d '{"UserOne": 1, "UserTwo": 2}'

{}

$> curl -X POST http://localhost:8000/api/match -d '{"UserOne": 1, "UserTwo": 3}'

{"error":"Unknown Users: 1 | 3"}

$> curl http://localhost:8000/api/match/1                          

{"ID":1,"UserOne":1,"UserTwo":2,"MatchedOn":"2020-02-19T12:00:53.721951Z"}
```

### DIFF
--- a/match/coms/service_coms_handler.go
+++ b/match/coms/service_coms_handler.go
@@ -17,8 +17,11 @@ func (coms *Handler) Init(config *utils.Config) {
 }
 
 // CheckUser makes requests to the target service to check if a userId exists.
-func (coms *Handler) CheckUser(service string, userID int) (bool, error) {
-	hostname := coms.Services[service]
+func (coms *Handler) CheckUser(userID int) (bool, error) {
+	hostname, ok := coms.Services["user"]
+	if !ok {
+		return false, fmt.Errorf("service %s's hostname not in config file", "user")
+	}
 
 	resp, err := http.Get(fmt.Sprintf("%s/%d", hostname, userID))
 	if err != nil {
--- a/match/coms/service_coms_handler.go
+++ b/match/coms/service_coms_handler.go
@@ -1,8 +1,8 @@
 package coms
 
 import (
-	"github.com/TempleEight/spec-golang/match/utils"
 	"fmt"
+	"github.com/TempleEight/spec-golang/match/utils"
 	"net/http"
 )
 
@@ -11,8 +11,8 @@ type Handler struct {
 	Services map[string]string
 }
 
-// Initialise sets up the Handler object with a list of services from the config
-func (coms *Handler) Initialise(config *utils.Config) {
+// Init sets up the Handler object with a list of services from the config
+func (coms *Handler) Init(config *utils.Config) {
 	coms.Services = config.Services
 }
 
--- a/match/coms/service_coms_handler.go
+++ b/match/coms/service_coms_handler.go
@@ -1,0 +1,30 @@
+package coms
+
+import (
+	"github.com/TempleEight/spec-golang/match/utils"
+	"fmt"
+	"net/http"
+)
+
+// Handler maintaints the list of services and their associated hostnames
+type Handler struct {
+	Services map[string]string
+}
+
+// Initialise sets up the Handler object with a list of services from the config
+func (coms *Handler) Initialise(config *utils.Config) {
+	coms.Services = config.Services
+}
+
+// CheckUser makes requests to the target service to check if a userId exists.
+func (coms *Handler) CheckUser(service string, userID int) (bool, error) {
+	hostname := coms.Services[service]
+
+	resp, err := http.Get(fmt.Sprintf("%s/%d", hostname, userID))
+	if err != nil {
+		return false, err
+	}
+	defer resp.Body.Close()
+
+	return resp.StatusCode == 200, nil
+}
--- a/match/config.json
+++ b/match/config.json
@@ -2,5 +2,8 @@
   "user": "postgres",
   "dbName": "postgres",
   "host": "match-db",
-  "sslMode": "disable"
+  "sslMode": "disable",
+  "services": {
+    "user": "http://user:80/user"
+  }
 }
--- a/match/match.go
+++ b/match/match.go
@@ -79,31 +79,28 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	userOne := *req.UserOne
-	userTwo := *req.UserTwo
-
-	userOneValid, err := coms.CheckUser(userOne)
+	userOneValid, err := coms.CheckUser(*req.UserOne)
 	if err != nil {
-		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "matches", err.Error()))
-		http.Error(w, errMsg, http.StatusBadRequest)
-		return
-	}
-
-	userTwoValid, err := coms.CheckUser(userTwo)
-	if err != nil {
-		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "matches", err.Error()))
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "user", err.Error()))
 		http.Error(w, errMsg, http.StatusBadRequest)
 		return
 	}
 
 	if !userOneValid {
-		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unknown User: %d", userOne))
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unknown User: %d", *req.UserOne))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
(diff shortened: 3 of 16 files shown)


match/match.go
```diff
@@ -73,6 +78,28 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+	userTwoValid, err := coms.CheckUser("user", userTwo)
+	if err != nil {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "matches", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	if !userOneValid || !userTwoValid {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unknown Users: %d | %d", userOne, userTwo))
```
c-harding: This is misleading if one is valid

match/coms/service_coms_handler.go
```diff
@@ -0,0 +1,30 @@
...
+	"net/http"
+)
+
+// Handler maintaints the list of services and their associated hostnames
+type Handler struct {
+	Services map[string]string
+}
+
+// Initialise sets up the Handler object with a list of services from the config
+func (coms *Handler) Initialise(config *utils.Config) {
```
c-harding: British English:


More to the point, is that what we want our generated code to be in?

match/match.go
jaylees14: Yeah, split up these errors I think

match/coms/service_coms_handler.go
jaylees14: Let's avoid it altogether with `init`?

match/match.go
```diff
@@ -94,8 +94,14 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !userOneValid || !userTwoValid {
-		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unknown Users: %d | %d", userOne, userTwo))
+	if !userOneValid {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unknown User: %d", userOne))
```
c-harding: Turning the question around, is this the error message we want when both are wrong? I guess it’s probably fine

c-harding (dismissed):

match/match.go
```diff
@@ -73,6 +78,34 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	userOne := *req.UserOne
+	userTwo := *req.UserTwo
```
jaylees14: I'm not sure about this, why don't we just dereference them when we need them?

jaylees14: @smithwjv What's your thoughts?

match/match.go
```diff
@@ -73,6 +78,34 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	userOneValid, err := coms.CheckUser("user", userOne)
+	if err != nil {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "matches", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
```
jaylees14, line 87: Whitespace after this line for consistency

match/coms/service_coms_handler.go
```diff
@@ -0,0 +1,30 @@
...
+	Services map[string]string
+}
+
+// Init sets up the Handler object with a list of services from the config
+func (coms *Handler) Init(config *utils.Config) {
+	coms.Services = config.Services
+}
+
+// CheckUser makes requests to the target service to check if a userId exists.
+func (coms *Handler) CheckUser(service string, userID int) (bool, error) {
```
jaylees14: Why are we passing the service name in here? Surely `CheckUser` can only be to one service

match/coms/service_coms_handler.go
```diff
@@ -0,0 +1,30 @@
...
+	hostname := coms.Services[service]
```
jaylees14: You need to check if the `service` key exists:
```
hostname, ok := coms.Services["user"]
if !ok {
  ...
}

File: match/coms/service_coms_handler.go (hunk 6 above)
Comment (jaylees14) [reply match/coms/service_coms_handler.go]: It also avoids the programmer of `matches.go` from knowing the service name

File: match/match.go (hunk 4 above)
Comment (Lewiky) [reply match/match.go]: I'm easy on this, motivation was to avoid dereferencing them twice each, but it doesn't save code

File: match/config.json [hunk 8]
```diff
@@ -2,5 +2,8 @@
   "user": "postgres",
   "dbName": "postgres",
   "host": "match-db",
-  "sslMode": "disable"
+  "sslMode": "disable",
```
c-harding, line 5: Reasons to dislike JSON #‌23: the lack of trailing commas

c-harding (approved):
All looks fine

match/coms/service_coms_handler.go
```diff
@@ -0,0 +1,33 @@
+package coms
```
willjvsmith: This filename uses `snake_case` when we've used `kebab-case` elsewhere.

match/coms/service_coms_handler.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	"github.com/TempleEight/spec-golang/match/utils"
+	"net/http"
+)
+
+// Handler maintaints the list of services and their associated hostnames
```
willjvsmith:
```suggestion
// Handler maintains the list of services and their associated hostnames
```

match/match.go
willjvsmith: Yeah I think we should dereference at use, as we pass the `req` object to `dao.CreateMatch`.

match/match.go
```diff
@@ -73,6 +78,35 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	userOne := *req.UserOne
+	userTwo := *req.UserTwo
+
+	userOneValid, err := coms.CheckUser(userOne)
+	if err != nil {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "matches", err.Error()))
```
willjvsmith:
```suggestion
		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "user", err.Error()))
```

match/match.go
```diff
@@ -73,6 +78,35 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	userTwoValid, err := coms.CheckUser(userTwo)
+	if err != nil {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "matches", err.Error()))
```
willjvsmith:
```suggestion
		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Unable to reach %s service: %s", "user", err.Error()))
```

match/match.go
```diff
@@ -73,6 +78,35 @@ func matchCreateHandler(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	if !userOneValid {
```
willjvsmith: I suggest moving this check to straight after the `checkUser` call for `userOne`, so the second `checkUser` call for `userTwo` isn't made if `userOne` isn't valid.

match/match.go
willjvsmith: I think it should fail on the first one, so we don't make a redundant `checkUser` call for `userTwo` when `userOne` is invalid anyway. See my comment below about shifting the calls and checks around.

willjvsmith (changes requested):
Few small things, looks pretty clean

willjvsmith (approved):
LGTM 🚀

jaylees14 (approved):

PR #14, opened 2020-02-20
[infra] Use go modules

jaylees14:
Use go modules for dependency management, following https://blog.golang.org/using-go-modules.

This has the advantage that it doesn't need to be in the GOPATH, and makes the `Dockerfile` much easier generate

### DIFF
--- a/match/Dockerfile
+++ b/match/Dockerfile
@@ -1,15 +1,12 @@
 FROM golang:1.13.7-alpine
 
-WORKDIR /go/src/github.com/TempleEight/spec-golang/match/
+WORKDIR /match
+
+COPY go.mod go.sum ./
+RUN go mod download
 
 COPY . .
 COPY config.json /etc/match-service/
-
-RUN apk add --no-cache git
-
-RUN go get github.com/gorilla/mux
-RUN go get github.com/lib/pq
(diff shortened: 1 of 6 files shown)


Lewiky (approved):

match/Dockerfile
```diff
@@ -1,16 +1,13 @@
...
 
-RUN apk add --no-cache git
-
-RUN go get github.com/gorilla/mux
-RUN go get github.com/lib/pq
-RUN go get github.com/asaskevich/govalidator
-
 RUN go build -o match
 
 ENTRYPOINT ./match
```
Lewiky, line 13: Can we get a new-line at the end of this file to keep in line with the other Dockerfile?

user/Dockerfile
```diff
@@ -1,18 +1,16 @@
...
-RUN go get github.com/lib/pq
-RUN go get github.com/asaskevich/govalidator
-
-RUN go build -o user
+RUN go build -o user 
 
 ENTRYPOINT ./user
 
 EXPOSE 80
+
```
c-harding:
```suggestion
```

c-harding (approved):
lgtm

PR #17, opened 2020-02-26
[auth] Add a basic auth service

jaylees14:
We need an auth service. This is where tokens will be issued based on an email/password authentication and then future services will require this token before giving access to data.

This PR simply adds a basic auth service, with Dockerfile / Postgres db etc. and hooks it in with Kong and Docker-Compose.

Test:
```
❯❯❯ docker-compose up --build
❯❯❯ sh kong/configure-kong.sh
❯❯❯ curl -X POST localhost:8000/api/auth
{}
```

### DIFF
--- a/auth-db/init.sql
+++ b/auth-db/init.sql
@@ -1,0 +1,5 @@
+CREATE TABLE auth (
+    id SERIAL PRIMARY KEY,
+    email TEXT,
+    password TEXT
+);
--- a/auth/Dockerfile
+++ b/auth/Dockerfile
@@ -1,0 +1,15 @@
+FROM golang:1.13.7-alpine
+
+WORKDIR /auth
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+COPY config.json /etc/auth-service/
+
+RUN go build -o auth
+
+ENTRYPOINT ./auth
+
+EXPOSE 82
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"log"
+	"net/http"
+
+	authDAO "github.com/TempleEight/spec-golang/auth/dao"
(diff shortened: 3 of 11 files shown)


auth/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN go mod download
+
+COPY . .
+COPY config.json /etc/auth-service/
+
+RUN go build -o auth
+
+ENTRYPOINT ./auth
+
+EXPOSE 82
```
willjvsmith, line 15: Do we need to be exposing different ports for each service?

willjvsmith (approved):

auth-db/init.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE auth (
+    id SERIAL PRIMARY KEY,
+    email TEXT,
+    password TEXT
```
c-harding, line 4: If this is hashed, isn’t there a constant size?

auth/Dockerfile
jaylees14: Good question: have done that here for consistency with the other services. I'll wait until we move to Kube to change it, if it is possible

auth/auth.go
```diff
@@ -0,0 +1,42 @@
...
+		log.Fatal(err)
+	}
+
+	r := mux.NewRouter()
+	r.HandleFunc("/auth", authCreateHandler).Methods(http.MethodPost)
+	log.Fatal(http.ListenAndServe(":82", r))
+}
+
+func authCreateHandler(w http.ResponseWriter, r *http.Request) {
+	json.NewEncoder(w).Encode(struct{}{})
```
c-harding, line 41: Security on point

c-harding (approved):

auth-db/init.sql
jaylees14: Good point. I'll leave this for now, then update when I've worked out what hashing we're using 😄

PR #18, opened 2020-02-26
[auth] Register a new email/password combination and return a JWT

jaylees14:
The first endpoint I'm creating is one for **creating** a new auth. You provide a email and password, and it should give you back a JWT, valid for 24 hours, associated with your email address. 

There's quite a lot of things that are *not* included yet - I'm aware of that - namely:
- The issuer and secret token (these will be derived from *Kong* - more to come on that)
- Refresh tokens
- Storing a "blacklist" of invalid JWTs
- The minor issue of it not being HTTPS 😉 

### Testing
Build the project in the usual way:
```
❯❯❯ docker-compose up --build
❯❯❯ sh kong/configure-kong.sh 
```

Create a new user:
```
❯❯❯ curl -X POST localhost:8000/api/auth -d '{"email": "jay@me.com", "password": "abcdefgh"}'
{"AccessToken":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImpheUBtZS5jb20iLCJleHAiOjE1ODI4MTYzNDIsImlzcyI6IlRPRE8ifQ.D3GbSqOO3_EXe4eW3gDxEIudwhvTZ-2-rp-Utvl1zaM"}
```

Validate the storage:
```
❯❯❯ docker exec -it spec-golang_auth-db_1 sh
...
# psql
...
postgres=# SELECT * FROM auth;
 id |   email    |                           password                           
----+------------+--------------------------------------------------------------
  1 | jay@me.com | $2a$10$7KdOXOKKcJzvy4QSGbL9B.G8iNanOPBC3Er3EY5bVoPFjnIuat1pq
(1 row)
```

Validate the JWT (note the +24 time in `exp`):
Screenshot 2020-02-26 at 15 14 11

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -84,7 +84,6 @@ func authCreateHandler(w http.ResponseWriter, r *http.Request) {
 		AccessToken: accessToken,
 	}
 	json.NewEncoder(w).Encode(response)
-
 }
 
 // Create an access token with a 24 hour lifetime
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -3,12 +3,16 @@ package main
 import (
 	"encoding/json"
 	"flag"
+	"fmt"
+	"golang.org/x/crypto/bcrypt"
 	"log"
 	"net/http"
+	"time"
 
 	authDAO "github.com/TempleEight/spec-golang/auth/dao"
 	"github.com/TempleEight/spec-golang/auth/utils"
 	valid "github.com/asaskevich/govalidator"
+	"github.com/dgrijalva/jwt-go"
 	"github.com/gorilla/mux"
 )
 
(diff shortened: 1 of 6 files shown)


auth/auth.go
```diff
@@ -38,5 +42,58 @@ func main() {
...
+
+	_, err = valid.ValidateStruct(req)
+	if err != nil {
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Invalid request parameters: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	// Hash and salt the password before storing
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(req.Password), bcrypt.DefaultCost)
```
Lewiky, line 61: Is the salt something implicit or does it need to be stored?

Lewiky (approved):
LGTM 

Are we agreeing on commiting the .mod and .sum files?

auth/auth.go
jaylees14: The salt is handled by the beautiful people who built `bcrypt`: 

https://github.com/golang/crypto/blob/2aa609cf4a9d7d1126360de73b55b6002f9e052a/bcrypt/bcrypt.go#L145

jaylees14:
@Lewiky yes - see here for justification: https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file

auth/auth.go
```diff
@@ -38,5 +42,58 @@ func main() {
...
+		errMsg := utils.CreateErrorJSON(fmt.Sprintf("Could not create access token: %s", err.Error()))
+		http.Error(w, errMsg, http.StatusInternalServerError)
+		return
+	}
+
+	response := authDAO.AuthCreateResponse{
+		AccessToken: accessToken,
+	}
+	json.NewEncoder(w).Encode(response)
+
```
willjvsmith: Remove extra line

auth/auth.go
```diff
@@ -38,5 +42,58 @@ func main() {
...
+}
+
+// Create an access token with a 24 hour lifetime
```
willjvsmith, line 89: Not something to be changed for this PR, but is this something we want to be customisable?

auth/dao/auth-dao.go
```diff
@@ -14,6 +14,16 @@ type DAO struct {
 	DB *sql.DB
 }
 
+// AuthCreateRequest contains the information required to create a new authorized user
+type AuthCreateRequest struct {
+	Email    string `valid:"email,required"`
+	Password string `valid:"type(string),required,stringlength(8|64)"`
```
willjvsmith, line 20: Similarly here.

In my opinion we don't allow them to customise it, but we make sure we have some literature to justify our choices. If they wanna change it they can alter the code.

auth/dao/auth-dao.go
```diff
@@ -24,3 +34,18 @@ func (dao *DAO) Init(config *utils.Config) error {
...
+	result, err := db.Exec(query, args...)
+	if err != nil {
+		return 0, err
+	}
+
+	return result.RowsAffected()
+}
+
+func (dao *DAO) CreateAuth(request AuthCreateRequest) error {
+	_, err := executeQuery(dao.DB, "INSERT INTO auth (email, password) VALUES ($1, $2) RETURNING *", request.Email, request.Password)
```
willjvsmith: Don't need the `RETURNING` clause

willjvsmith (approved):

auth/dao/auth-dao.go
jaylees14: Sounds good. I'll address this in a follow up PR that focusses on setting these limits to something other than arbitrary values.

PR #20, opened 2020-02-27
[auth] Request JWT secret from Kong

jaylees14:
Rather than managing secret keys to sign JWTs ourselves, we can leverage additional Kong functionality. This has several advantages:
- Kong will automatically require and validate JWTs before calling any other service
- Kong controls the secure stuff, so we don't have to
- Less code ❤️ 

So... to do this, the Auth service has to do several things (followed from https://docs.konghq.com/hub/kong-inc/jwt/)
- Register as a "consumer"
- Use this registration to request a secret & key to sign JWTs with

To do this, the auth service requires access to Kong's "Admin" port, exposed on 8001. This would normally _not_ be exposed to the general public, however we currently have it configured that way for ease of testing / configuration.

One issue we now have is that the auth service depends on the Kong service already being ready and active. This **isn't** possible from the Docker Compose `depends-on`(https://docs.docker.com/compose/compose-file/#depends_on), so we are forced to use a shell script that periodically checks for readiness, following the advice from https://docs.docker.com/compose/startup-order/. 

### Test Plan
First spin up all the containers and configure Kong:
```
 ❯❯❯ docker-compose up --build  
 ❯❯❯ sh kong/configure-kong.sh
```

Now try and make a request to a service that isn't auth:
```
❯❯❯ curl -X GET localhost:8000/api/user/2
{"message":"Unauthorized"}
```

Now create a new user:
```
❯❯❯ curl -X POST localhost:8000/api/auth -d '{"email":"jay@test.com", "password":"abcdefgh"}'
{"AccessToken":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImpheUB0ZXN0LmNvbSIsImV4cCI6MTU4Mjg5MzU3NywiaXNzIjoiMG9kVHc1d1FXOHRVdlUxVEM5elQxa2pjNnpKQVM1RVYifQ.s89vHXXsZ2xpiCdMGtpABcRA6kbCIlgCWxV3v9SekAI"}
```

Now make the same request as before, including the token in the headers - it works 😍 
```
❯❯❯ curl -X GET localhost:8000/api/user/2 -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImpheUB0ZXN0LmNvbSIsImV4cCI6MTU4Mjg5MzU3NywiaXNzIjoiMG9kVHc1d1FXOHRVdlUxVEM5elQxa2pjNnpKQVM1RVYifQ.s89vHXXsZ2xpiCdMGtpABcRA6kbCIlgCWxV3v9SekAI"
{"error":"user not found with ID 2"}
```

Now to make sure this is actually checking the JWT correctly, I've created two JWTs using the debugger at [jwt.io](jwt.io) - one that contains the correct `iss` tag for the `auth-service`, but incorrect signature, and one with random contents:

#### Correct `iss`, incorrect signature
```
❯❯❯ curl -X GET localhost:8000/api/user/2 -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJpc3MiOiIwb2RUdzV3UVc4dFV2VTFUQzl6VDFramM2ekpBUzVFViJ9.1bliwzt-ZllhodP0anzB94htvuFsDcykhzeuV_jRlNE"
{"message":"Invalid signature"}
```
Screenshot 2020-02-27 at 12 45 12

#### Random Contents:
```
 ❯❯❯ curl -X GET localhost:8000/api/user/2 -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJpc3MiOiJSQU5ET01QRVJTT04ifQ._Nv1aKd18JqKf6JVIeUd2-IWNFlPF2pURGRdN19UI1Y"
{"message":"No credentials found for given 'iss'"}
```
Screenshot 2020-02-27 at 12 43 24

### DIFF
--- a/auth/comm/handler.go
+++ b/auth/comm/handler.go
@@ -42,7 +42,7 @@ func createKongConsumer(hostname string) (*consumerResponse, error) {
 	}
 	defer res.Body.Close()
 
-	if res.StatusCode != 201 {
+	if res.StatusCode != http.StatusCreated {
 		// If we have an error code, the message _should_ be in the body
 		bodyBytes, err := ioutil.ReadAll(res.Body)
 		if err != nil {
@@ -68,7 +68,7 @@ func requestCredential(hostname string, consumer *consumerResponse) (*JWTCredent
 	}
 	defer res.Body.Close()
 
-	if res.StatusCode != 201 {
+	if res.StatusCode != http.StatusCreated {
 		// If we have an error code, the message _should_ be in the body
 		bodyBytes, err := ioutil.ReadAll(res.Body)
 		if err != nil {
--- a/auth/comm/handler.go
+++ b/auth/comm/handler.go
@@ -1,0 +1,104 @@
+package comm
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/TempleEight/spec-golang/auth/utils"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+)
+
+// Handler maintains the list of services and their associated hostnames
+type Handler struct {
+	Services map[string]string
+}
+
+// consumerResponse encapsulates the response from Kong after creating a consumer
+type consumerResponse struct {
+	ID       string `json:"id"`
+	Username string `json:"username"`
+}
+
+// JWTCredential store the issuer and secret key that must be used to sign requests
+type JWTCredential struct {
+	Key    string `json:"key"`
+	Secret string `json:"secret"`
+}
+
+// Init sets up the Handler object with a list of services from the config
+func (coms *Handler) Init(config *utils.Config) {
+	coms.Services = config.Services
+}
+
+func createKongConsumer(hostname string) (*consumerResponse, error) {
+	postData := url.Values{}
+	postData.Set("username", "auth-service")
+
+	res, err := http.PostForm(fmt.Sprintf("%s/consumers", hostname), postData)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != 201 {
+		// If we have an error code, the message _should_ be in the body
+		bodyBytes, err := ioutil.ReadAll(res.Body)
+		if err != nil {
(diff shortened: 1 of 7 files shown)


auth/comm/handler.go
```diff
@@ -0,0 +1,104 @@
...
+
+func requestCredential(hostname string, consumer *consumerResponse) (*JWTCredential, error) {
+	reqUrl := fmt.Sprintf("%s/consumers/%s/jwt", hostname, consumer.Username)
+	res, err := http.Post(reqUrl, "application/x-www-form-urlencoded", nil)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != 201 {
```
willjvsmith: Perhaps replace `201` with the http enum

auth/comm/handler.go
```diff
@@ -0,0 +1,104 @@
...
+	postData := url.Values{}
+	postData.Set("username", "auth-service")
+
+	res, err := http.PostForm(fmt.Sprintf("%s/consumers", hostname), postData)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != 201 {
```
willjvsmith: Same here

willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
YAS QUEEN LGTM BOI Espanol Espanol Numbah One

PR #21, opened 2020-02-27
[auth] Allow existing users to login and receive a *fresh* auth token

jaylees14:
Allow existing users who have registered with the service to login and get a new auth token, valid for 24 hours.

#### Test Plan
Configure the system in the standard way:
```
❯❯❯ docker compose up --build
❯❯❯ sh kong/configure-kong.sh
```

Make a request to the auth server with an non-existent login:
```
❯❯❯ curl -X GET localhost:8000/api/auth -d '{"email":"jay@me.com", "password":"abcdefgh"}'
{"error":"Invalid email or password"}
```

Register the user:
```
❯❯❯ curl -X POST localhost:8000/api/auth -d '{"email":"jay@me.com", "password":"abcdefgh"}'
{"AccessToken":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImpheUBtZS5jb20iLCJleHAiOjE1ODI5MDQ0MTgsImlzcyI6IlRPRE8ifQ.7AY4BPgEKhBgXOaeWEYmIgp_GrOJdafY0I5ZSmPJXL8"}
```

Login with the user:
```
❯❯❯ curl -X GET localhost:8000/api/auth -d '{"email":"jay@me.com", "password":"abcdefgh"}'
{"AccessToken":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImpheUBtZS5jb20iLCJleHAiOjE1ODI5MDQ0MzUsImlzcyI6IlRPRE8ifQ.gj7AKbhBa2pk54yhBYlU9u1-8bO9z1BdKT2jeYRVDpo"}
```

Login with incorrect password:
```
❯❯❯ curl -X GET localhost:8000/api/auth -d '{"email":"jay@me.com", "password":"abcdefghijklmn"}'
{"error":"Invalid email or password"}
```

### DIFF
--- a/auth/dao/auth-dao.go
+++ b/auth/dao/auth-dao.go
@@ -15,24 +15,24 @@ type DAO struct {
 }
 
 // CreateAuthRequest contains the information retrieved from an end user to create a new auth
-type CreateAuthRequest struct {
+type AuthCreateRequest struct {
 	Email    string `valid:"email,required"`
 	Password string `valid:"type(string),required,stringlength(8|64)"`
 }
 
 // ReadAuthRequest contains the information retrieved from an end user to validate an existing auth
-type ReadAuthRequest struct {
+type AuthReadRequest struct {
 	Email    string `valid:"email,required"`
 	Password string `valid:"type(string),required,stringlength(8|64)"`
 }
 
 // CreateAuthResponse contains an access token associated to a given auth
-type CreateAuthResponse struct {
+type AuthCreateResponse struct {
 	AccessToken string
 }
 
 // ReadAuthResponse contains an access token associated to a given auth
-type ReadAuthResponse struct {
+type AuthReadResponse struct {
 	AccessToken string
 }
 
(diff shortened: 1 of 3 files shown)


auth/dao/auth-dao.go
```diff
@@ -14,16 +14,24 @@ type DAO struct {
 	DB *sql.DB
 }
 
-// AuthCreateRequest contains the information required to create a new authorized user
-type AuthCreateRequest struct {
+// PlaintextAuth contains the information retrieved from an end user
+type PlaintextAuth struct {
```
jaylees14: @smithwjv What do you think about these types? I've gone away from the `*Request` and `*Response` types because they're always the same - but distinguished between plaintext and hashed so that it's clear what types we're storing/retrieving?

auth/dao/errors.go
```diff
@@ -0,0 +1,6 @@
+package dao
+
+import "errors"
+
+// ErrAuthNotFound is returned when an for the provided email was not found
```
willjvsmith: Typo

Lewiky (approved):

auth/dao/auth-dao.go
willjvsmith: I think I marginally prefer sticking to the request response paradigm and commenting the structs with whether they're plaintext or hashed, just to be inline with everything else and not take longer to read. No particularly strong feelings though.

willjvsmith: Though I guess you could call them `PlaintextAuthCreateRequest` and `HashedAuthCreateResponse`.

auth/dao/auth-dao.go
```diff
@@ -45,7 +53,29 @@ func executeQuery(db *sql.DB, query string, args ...interface{}) (int64, error)
...
+}
+
+// CreateAuth persists a new auth'd user to the data store
+func (dao *DAO) CreateAuth(request HashedAuth) error {
 	_, err := executeQuery(dao.DB, "INSERT INTO auth (email, password) VALUES ($1, $2)", request.Email, request.Password)
 	return err
 }
+
+// GetAuth attempts to find an existing auth'd user in the data store
+func (dao *DAO) GetAuth(request PlaintextAuth) (*HashedAuth, error) {
```
willjvsmith: What do you think about calling these `gets` `reads`? Just to stick to CRUD

willjvsmith (approved):

PR #31, opened 2020-03-03
[user] Generalise database types to allow for dependency injection

jaylees14:
To fully test this service, we need to be a bit more general about the code that's being written.

Now, we define a generic interface `Datastore`, which is implemented by the `DAO` object.
This means the main class can be defined over methods on this datastore, allowing dependency injection for testing.

This shouldn't make any difference to existing functionality, so previous test plans still apply:

```
❯❯❯ docker-compose up --build
❯❯❯ sh kong/configure-kong.sh
```

```
❯❯❯ ACCESS=$(curl -s -X POST localhost:8000/api/auth -d '{"email":"jay@test.com", "password":"abcdefgh"}' | jq -r .AccessToken)
❯❯❯ USER1=$(curl -s -X POST localhost:8000/api/user -d '{"name": "Jay"}' -H "Authorization: Bearer $ACCESS" | jq -r .ID)
❯❯❯ USER2=$(curl -s -X POST localhost:8000/api/user -d '{"name": "Lewis"}' -H "Authorization: Bearer $ACCESS" | jq -r .ID)
❯❯❯ curl -X POST localhost:8000/api/match -d "{\"userone\": $USER1, \"usertwo\": $USER2}" -H "Authorization: Bearer $ACCESS"
```

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -13,7 +13,9 @@ import (
 	"github.com/gorilla/mux"
 )
 
-var dao userDAO.DAO
+type Env struct {
+	dao userDAO.Datastore
+}
 
 func main() {
 	configPtr := flag.String("config", "/etc/user-service/config.json", "configuration filepath")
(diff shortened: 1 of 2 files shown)


user/user.go
```diff
@@ -13,7 +13,9 @@ import (
 	"github.com/gorilla/mux"
 )
 
-var dao userDAO.DAO
+type Env struct {
+	dao userDAO.Datastore
```
willjvsmith: Could rename to just `dao` now, which would be nice

willjvsmith (approved):
Trust it works until we add tests with this 🚀 

Nice one

c-harding (approved):

PR #36, opened 2020-03-03
[k8s] Integrate Kong and Services for end to end K8s config

Lewiky:
The last of the Kubernooties PRs - _Tying everything together_.

This adds the authentication service to kubernetes via the same methods used with the user and match services (literally remove the word 'user' and replace with the word 'auth' - very good for generation)

Also provides an end to end build script that sorts out all of the kong configuration to forward requests to the correct services and sets environment variables for the kubernetes endpoints.

Also this doesn't break any of the docker-compose stuff if you don't feel like running the whole k8s stack for future development.

## Test plan

With Minikube installed as before: 

#### Set environment vars

Set the: 
- `$REG_URL`
- `$REG_USERNAME`
- `$REG_PASSWORD`
- `$REG_EMAIL`

Environment variables as before - (NB: This could be edited out by moving the images to docker-hub?)

Run: 

```
source deploy.sh
```

Wait for this to finish, it'll make sure everything is configured correctly

```
❯❯❯ ACCESS=$(curl -s -X POST $KONG_ENTRY/api/auth -d '{"email":"jay@test.com", "password":"abcdefgh"}' | jq -r .AccessToken)
❯❯❯ USER1=$(curl -s -X POST $KONG_ENTRY/api/user -d '{"name": "Jay"}' -H "Authorization: Bearer $ACCESS" | jq -r .ID)
❯❯❯ USER2=$(curl -s -X POST $KONG_ENTRY/api/user -d '{"name": "Lewis"}' -H "Authorization: Bearer $ACCESS" | jq -r .ID)
❯❯❯ curl -X POST $KONG_ENTRY/api/match -d "{\"userone\": $USER1, \"usertwo\": $USER2}" -H "Authorization: Bearer $ACCESS"
```

### DIFF
--- a/K8s-DEPLOYMENT.md
+++ b/K8s-DEPLOYMENT.md
@@ -1,6 +1,6 @@
 # Kubernetes Deployment
 
-#### _NEW: Automated Deployment!_
+## _NEW:_ Automated Deployment
 _Simply run `source deploy.sh` to do all these steps for you!_
 
 Note: When finished with the cluster, you still need to clean up with
--- a/K8s-DEPLOYMENT.md
+++ b/K8s-DEPLOYMENT.md
@@ -3,6 +3,14 @@
 #### _NEW: Automated Deployment!_
 _Simply run `source deploy.sh` to do all these steps for you!_
 
+Note: When finished with the cluster, you still need to clean up with
+
+```
+$ minikube delete
+```
+
+## Manual Deployment
+
 To deploy this project onto a kubernetes cluster follow these steps:
 
 ## Cluster setup
--- a/K8s-DEPLOYMENT.md
+++ b/K8s-DEPLOYMENT.md
@@ -1,5 +1,8 @@
 # Kubernetes Deployment
 
+> __ NEW: Automated Deployment!  __
+> Simply run `source deploy.sh` to do all these steps for you!
+
 To deploy this project onto a kubernetes cluster follow these steps:
 
 ## Cluster setup
--- a/K8s-DEPLOYMENT.md
+++ b/K8s-DEPLOYMENT.md
@@ -1,7 +1,7 @@
 # Kubernetes Deployment
 
-> __ NEW: Automated Deployment!  __
-> Simply run `source deploy.sh` to do all these steps for you!
+#### _NEW: Automated Deployment!_
+_Simply run `source deploy.sh` to do all these steps for you!_
 
 To deploy this project onto a kubernetes cluster follow these steps:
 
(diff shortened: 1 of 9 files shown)


K8s-DEPLOYMENT.md
```diff
@@ -1,5 +1,8 @@
 # Kubernetes Deployment
 
+#### _NEW: Automated Deployment!_
+_Simply run `source deploy.sh` to do all these steps for you!_
```
c-harding: This should be more isolated, with more clarity on which bits below still apply (cleanup etc)

deploy.sh
```diff
@@ -0,0 +1,48 @@
+#! /bin/bash
+# Deployment script for kube - generates configmaps from SQL init files and provisions everything
+
+GREEN="\033[1;32m"
+BLUE="\033[1;34m"
+NOCOLOR="\033[0m"
+
+minikube start --vm-driver=virtualbox
+
+echo ${GREEN}
```
c-harding: What are these braces for?

deploy.sh
```diff
@@ -0,0 +1,48 @@
+#! /bin/bash
```
jaylees14: Do we explicitly need `bash`? The other `.sh` script uses `#!/bin/sh`

deploy.sh
```diff
@@ -0,0 +1,48 @@
...
+echo ${GREEN}
+
+kubectl create secret docker-registry regcred --docker-server=$REG_URL --docker-username=$REG_USERNAME --docker-password=$REG_PASSWORD --docker-email=$REG_EMAIL
+kubectl create configmap match-db-config --from-file match-db/init.sql -o=yaml
+kubectl create configmap user-db-config --from-file user-db/init.sql -o=yaml
+kubectl create configmap auth-db-config --from-file auth-db/init.sql -o=yaml
+kubectl create -f kube/kong
+kubectl create -f kube/user
+kubectl create -f kube/match
+kubectl create -f kube/auth
```
jaylees14: `ls -d kube/* | xargs -I"@" kubectl create -f @` if you prefer 😉

deploy.sh
```diff
@@ -0,0 +1,48 @@
...
+echo ${BLUE}
+
+echo KONG_ENTRY: $KONG_ENTRY
+echo KONG_ADMIN: $KONG_ADMIN
+
+echo ${NOCOLOR}
+
+echo Configuring Kong...
+
+sh kong/configure-kong-k8s.sh
```
jaylees14, line 53: Maybe just a little echo after this to say done?

deploy.sh
jaylees14: Jk this probably isn't as readable

jaylees14 (approved):
LGTM, few minor things aside from @xsanda's comments

deploy.sh
Lewiky: I'll do this as a for loop instead

K8s-DEPLOYMENT.md
```diff
@@ -3,6 +3,14 @@
 #### _NEW: Automated Deployment!_
```
c-harding:
```suggestion
## _NEW:_ Automated Deployment
```
This emphasises its hierarchical equivalence with manual deployment

c-harding (approved):

PR #39, opened 2020-03-03
[API] Add the entire match service to OpenAPI

c-harding:
Lots of repetition to mirror the structure of

https://github.com/TempleEight/spec-golang/blob/develop/match/dao/dao.go#L27


jaylees14 (approved):

Lewiky (approved):
Dreamy

PR #41, opened 2020-03-04
[API] Add the user service to the OpenAPI spec

c-harding:
This is branched off #39, so the diff will shrink once it is merged. Basically the same thing as before, but with some field and string substitutions.


jaylees14 (approved):
LGTM

willjvsmith (approved):
LGTM

PR #43, opened 2020-03-05
[match][test] Add unit test basis for match service

jaylees14:
Add basis for unit tests in match service.

We hit all the cases in `create` except for internal server errors and `govalidator` checks - since there are no conditions on the valiadator
```
go test -coverprofile=test.html && go tool cover -html=test.html
```

### DIFF
--- a/match/match.go
+++ b/match/match.go
@@ -19,6 +19,18 @@ type Env struct {
 	comm comm.Comm
 }
 
+func Router(env Env) *mux.Router {
+	r := mux.NewRouter()
+	// Mux directs to first matching route, i.e. the order matters
+	r.HandleFunc("/match/all", env.matchListHandler).Methods(http.MethodGet)
+	r.HandleFunc("/match", env.matchCreateHandler).Methods(http.MethodPost)
+	r.HandleFunc("/match/{id}", env.matchReadHandler).Methods(http.MethodGet)
+	r.HandleFunc("/match/{id}", env.matchUpdateHandler).Methods(http.MethodPut)
+	r.HandleFunc("/match/{id}", env.matchDeleteHandler).Methods(http.MethodDelete)
+	r.Use(jsonMiddleware)
+	return r
+}
+
 func main() {
 	configPtr := flag.String("config", "/etc/match-service/config.json", "configuration filepath")
 	flag.Parse()
@@ -39,15 +51,7 @@ func main() {
 
 	env := Env{d, c}
 
-	r := mux.NewRouter()
-	// Mux directs to first matching route, i.e. the order matters
-	r.HandleFunc("/match/all", env.matchListHandler).Methods(http.MethodGet)
-	r.HandleFunc("/match", env.matchCreateHandler).Methods(http.MethodPost)
-	r.HandleFunc("/match/{id}", env.matchReadHandler).Methods(http.MethodGet)
-	r.HandleFunc("/match/{id}", env.matchUpdateHandler).Methods(http.MethodPut)
-	r.HandleFunc("/match/{id}", env.matchDeleteHandler).Methods(http.MethodDelete)
-	r.Use(jsonMiddleware)
-
+	r := Router(env)
 	log.Fatal(http.ListenAndServe(":81", r))
 }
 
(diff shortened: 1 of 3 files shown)


match/match_test.go
```diff
@@ -0,0 +1,202 @@
...
+type MockDAO struct {
+	Matches []MockMatch
+}
+
+type MockComm struct {
+	UserIds []int
+}
+
+func (md *MockDAO) ListMatch() (*dao.MatchListResponse, error) {
+	matches := make([]dao.MatchReadResponse, 0)
```
willjvsmith: We wanna avoid plurals, think we've used `MatchList` in place

match/match_test.go
```diff
@@ -0,0 +1,202 @@
...
+	UserTwo   int
+	MatchedOn string
+}
+
+type MockDAO struct {
+	Matches []MockMatch
+}
+
+type MockComm struct {
+	UserIds []int
```
willjvsmith:
```suggestion
	UserIDs []int
```

willjvsmith: smh

match/match_test.go
```diff
@@ -0,0 +1,202 @@
...
+	rec := httptest.NewRecorder()
+	req, err := http.NewRequest(method, url, strings.NewReader(body))
+	if err != nil {
+		return nil, err
+	}
+	Router(env).ServeHTTP(rec, req)
+	return rec, nil
+}
+
+// Test that a match can be created successfully, assuming the users exist
```
willjvsmith: Hard, but need to avoid plurals in comments too: `users`

match/match_test.go
```diff
@@ -0,0 +1,202 @@
...
+		t.Fatalf("Could not make request: %s", err.Error())
+	}
+
+	if res.Code != http.StatusBadRequest {
+		t.Errorf("Wrong status code %v", res.Code)
+	}
+}
+
+// Test that a match is not created if both of the users don't exist
+func TestMatchCreateHandlerFailsOnBothInvalidUsers(t *testing.T) {
```
willjvsmith: `Both` isn't very generation friendly, could have three users or four

c-harding (approved):
Nice

match/match_test.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		t.Fatalf("Could not make request: %s", err.Error())
+	}
+
+	if res.Code != http.StatusBadRequest {
+		t.Errorf("Wrong status code %v", res.Code)
+	}
+}
+
+func TestMatchCreateHandlerFailsOnOnlyProvidingOneUser(t *testing.T) {
```
willjvsmith, line 189: Me looking for the comment

willjvsmith (commented):
Code is 💯, but gotta be careful on naming stuff keeping generation in mind

match/match.go
```diff
@@ -39,15 +51,7 @@ func main() {
...
-	// Mux directs to first matching route, i.e. the order matters
-	r.HandleFunc("/match/all", env.matchListHandler).Methods(http.MethodGet)
-	r.HandleFunc("/match", env.matchCreateHandler).Methods(http.MethodPost)
-	r.HandleFunc("/match/{id}", env.matchReadHandler).Methods(http.MethodGet)
-	r.HandleFunc("/match/{id}", env.matchUpdateHandler).Methods(http.MethodPut)
-	r.HandleFunc("/match/{id}", env.matchDeleteHandler).Methods(http.MethodDelete)
-	r.Use(jsonMiddleware)
-
+	r := Router(env)
 	log.Fatal(http.ListenAndServe(":81", r))
```
willjvsmith:
```suggestion
	log.Fatal(http.ListenAndServe(":81", Router(env)))
```

willjvsmith: I swear to god

match/match_test.go
c-harding: `user parameters`?

willjvsmith (approved):

willjvsmith:
Little bit worried about the assumptions being made, i.e. every user, when theoretically a mix of fields could be being sent

PR #44, opened 2020-03-05
[auth][test] Add unit test basis for auth service

jaylees14:
Wow such verbosity.

Add all the boilerplate we need to write tests for the auth service.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/TempleEight/spec-golang/auth/comm"
+	"github.com/TempleEight/spec-golang/auth/dao"
+	"github.com/dgrijalva/jwt-go"
+)
+
+type MockAuth struct {
+	ID       int
+	Email    string
+	Password string
+}
+
+type MockDAO struct {
+	AuthList []MockAuth
+}
+
+type MockComm struct{}
+
+func (md *MockDAO) CreateAuth(request dao.AuthCreateRequest) (*dao.Auth, error) {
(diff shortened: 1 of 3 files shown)


auth/auth_test.go
```diff
@@ -0,0 +1,125 @@
...
+	if !ok {
+		t.Fatalf("Token doesn't contain an access token: %s", err.Error())
+	}
+
+	token, _, err := new(jwt.Parser).ParseUnverified(rawToken, jwt.MapClaims{})
+	if err != nil {
+		t.Fatalf("Could not decode JWT: %s", err.Error())
+	}
+
+	claims, ok := token.Claims.(jwt.MapClaims)
```
c-harding, line 103: What is this syntax?

c-harding (approved):

auth/auth_test.go
jaylees14: https://golang.org/ref/spec#Type_assertions

Lewiky (approved):

auth/go.mod
```diff
@@ -4,8 +4,8 @@ go 1.13
 
 require (
 	github.com/asaskevich/govalidator v0.0.0-20200108200545-475eaeb16496
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible // indirect
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
```
willjvsmith, line 7: What's the `incompatible` referring to?

willjvsmith (approved):

auth/go.mod
jaylees14: https://github.com/dgrijalva/jwt-go/issues/294 Looks like it's a `go mod` thing, where `jwt-go` doesn't fully support modules in v3, but will do in v4. Looks like it's in preview for now, but we can upgrade if/when they release, although they mention will be some breaking changes

PR #45, opened 2020-03-06
[auth][test] Add additional tests for auth

jaylees14:
I just <3 testing. 

Adds some more tests for the `auth` service. Follows the same structure as those for `user`.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -124,7 +124,7 @@ func TestAuthCreateHandlerSucceeds(t *testing.T) {
 	}
 }
 
-// Test that providing an empty value for the `Email` parameter causes a `StatusBadRequest`
+// Test that providing an empty value for the `email` parameter causes a `StatusBadRequest`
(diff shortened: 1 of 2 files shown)


auth/auth_test.go
```diff
@@ -123,3 +123,170 @@ func TestAuthCreateHandlerSucceeds(t *testing.T) {
...
+		&mockComm,
+		cred,
+	}
+
+	res, err := makeRequest(mockEnv, http.MethodGet, "/auth", `{"email": "idonotexist@test.com", "password":"BlackcurrantCrush123"}`)
+	if err != nil {
+		t.Fatalf("Could not make GET request: %s", err.Error())
+	}
+
+	if res.Code != http.StatusUnauthorized {
```
jaylees14, line 269: @xsanda Do you think this makes sense? Provide an email that doesn't exist, we return `Unauthorized`? Could also go with `BadRequest` maybe?

c-harding: lgtm

Lewiky (approved):

c-harding (approved):

PR #47, opened 2020-03-06
[user][dao] Move request response types from DAO

jaylees14:
So that we can easily modify the `request` and `response` types using `@serverSet` values etc. we have separated these from the actual types that are stored in the datastore.

This also means the DAO is completely separate from request / response - following the Go structure of isolation by responsibility.

All unit tests should still pass since the actual content of requests/responses shouldn't change.

### DIFF
--- a/user/dao/dao.go
+++ b/user/dao/dao.go
@@ -11,10 +11,10 @@ import (
 
 // Datastore provides the interface adopted by the DAO, allowing for mocking
 type Datastore interface {
-	CreateUser(request UserCreateRequest) (*UserCreateResponse, error)
-	ReadUser(userID int64) (*UserReadResponse, error)
-	UpdateUser(userID int64, request UserUpdateRequest) (*UserUpdateResponse, error)
-	DeleteUser(userID int64) error
+	CreateUser(input CreateUserInput) (*User, error)
+	ReadUser(input ReadUserInput) (*User, error)
+	UpdateUser(input UpdateUserInput) (*User, error)
+	DeleteUser(input DeleteUserInput) error
 }
 
 // DAO encapsulates access to the database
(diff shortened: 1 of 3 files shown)


user/dao/dao.go
```diff
@@ -11,43 +11,42 @@ import (
...
+// User encapsulates the object stored in the datastore
+type User struct {
+	ID   int64
+	Name string
 }
 
-// UserUpdateRequest contains all the information about an existing user
-type UserUpdateRequest struct {
-	Name string `valid:"type(string),required,stringlength(2|255)"`
+// CreateUserInput encapsulates the information required to create a user
```
c-harding: `a user`/`an user` is awkward to generate

user/dao/dao.go
```diff
@@ -11,43 +11,42 @@ import (
...
+type CreateUserInput struct {
+	Name string
 }
 
-// UserCreateResponse contains the information about the newly created user
-type UserCreateResponse struct {
-	ID   int
-	Name string
+// ReadUserInput encapsulates the information required to read a user
```
c-harding: `to specify which user to fetch`

user/dao/dao.go
```diff
@@ -11,43 +11,42 @@ import (
...
+type ReadUserInput struct {
+	ID int64
 }
 
-// UserReadResponse returns all the information stored about a user
-type UserReadResponse struct {
-	ID   int
+// UpdateUserInput encapsulates the information required to update a user
```
c-harding: `a single user`?

c-harding (commented):
Many occurences of `a user`, which is awkward to generate

Lewiky (approved):

user/dao/dao.go
```diff
@@ -76,10 +75,10 @@ func Init(config *util.Config) (*DAO, error) {
 }
 
 // CreateUser creates a new user in the database, returning the newly created user
-func (dao *DAO) CreateUser(request UserCreateRequest) (*UserCreateResponse, error) {
-	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO User_Temple (name) VALUES ($1) RETURNING *", request.Name)
+func (dao *DAO) CreateUser(input CreateUserInput) (*User, error) {
+	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO User_Temple (name) VALUES ($1) RETURNING *", input.Name)
 
-	var resp UserCreateResponse
+	var resp User
```
willjvsmith: Since in this pattern we're called the DAO types, `input`, and the handler types `request` and `response`, we should call this `output`, or `user`

user/user.go
```diff
@@ -13,17 +13,46 @@ import (
...
+	Name string `valid:"type(string),required,stringlength(2|255)"`
+}
+
+// createUserResponse contains the information about the newly created user
+type createUserResponse struct {
+	ID   int64
+	Name string
+}
+
+// readUserResponse returns all the information stored about a user
```
willjvsmith:
```suggestion
// readUserResponse returns all the information stored about a single user
```

willjvsmith (approved):

PR #48, opened 2020-03-07
[user][test] Make tests consistent between endpoints

jaylees14:
- Move tests to the structure `<verb><service><type>`
- Make comments consistent between each test
- Make ordering of tests consistent between handlers
- Avoid names where we require the use of `a` or `an` depending on the name

### DIFF
--- a/user/user_test.go
+++ b/user/user_test.go
@@ -77,7 +77,7 @@ func TestCreateUserHandlerSucceeds(t *testing.T) {
 		&MockDAO{UserList: make([]dao.User, 0)},
 	}
 
-	// Make a single user
+	// Create a single user
 	res, err := makeRequest(mockEnv, http.MethodPost, "/user", `{"Name": "Jay"}`)
 	if err != nil {
 		t.Fatalf("Could not make request: %s", err.Error())
@@ -100,7 +100,7 @@ func TestCreateUserHandlerFailsOnEmptyParameter(t *testing.T) {
 		&MockDAO{UserList: make([]dao.User, 0)},
 	}
 
-	// Make a single user
+	// Create a single user
 	res, err := makeRequest(mockEnv, http.MethodPost, "/user", `{"Name": ""}`)
 	if err != nil {
 		t.Fatalf("Could not make request: %s", err.Error())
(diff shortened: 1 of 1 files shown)


Lewiky (approved):

user/user_test.go
```diff
@@ -127,19 +147,19 @@ func TestUserCreateHandlerFailsOnNoBody(t *testing.T) {
...
 
-	// Make a user
+	// Make a single user
 	_, err := makeRequest(mockEnv, http.MethodPost, "/user", `{"Name": "Jay"}`)
 	if err != nil {
 		t.Fatalf("Could not make POST request: %s", err.Error())
 	}
 
-	// Access that same user
+	// Read that same user
```
willjvsmith: Not a big deal, but would be nice if `// Create a single user`

user/user_test.go
```diff
@@ -173,8 +193,8 @@ func TestUserReadHandlerFailsOnEmptyID(t *testing.T) {
 	}
 }
 
-// Test that providing an invalid ID to the read endpoint fails
-func TestUserReadHandlerFailsOnNonExistentID(t *testing.T) {
+// Test that providing a non existent ID to the read endpoint fails
```
willjvsmith:
```suggestion
// Test that providing a non-existent ID to the read endpoint fails
```

user/user_test.go
```diff
@@ -293,14 +336,14 @@ func TestUserUpdateHandlerFailsOnEmptyID(t *testing.T) {
...
-	// 404 Not Found, since no route is defined for PUT at /user
+	// Not Found, since no route is defined for PUT at /user
 	if res.Code != http.StatusNotFound {
 		t.Errorf("Wrong status code %v", res.Code)
 	}
 }
 
-// Test that providing an invalid ID to the update endpoint fails
-func TestUserUpdateHandlerFailsOnNonExistentID(t *testing.T) {
+// Test that providing a non existent ID to the update endpoint fails
```
willjvsmith:
```suggestion
// Test that providing a non-existent ID to the update endpoint fails
```

user/user_test.go
```diff
@@ -373,14 +416,14 @@ func TestUserDeleteHandlerFailsOnEmptyID(t *testing.T) {
...
-	// 404 Not Found, since no route is defined for DELETE at /user
+	// Not Found, since no route is defined for DELETE at /user
 	if res.Code != http.StatusNotFound {
 		t.Errorf("Wrong status code %v", res.Code)
 	}
 }
 
-// Test that providing an invalid ID to the delete endpoint fails
-func TestUserDeleteHandlerFailsOnNonExistentID(t *testing.T) {
+// Test that providing a non existent ID to the delete endpoint fails
```
willjvsmith:
```suggestion
// Test that providing a non-existent ID to the delete endpoint fails
```

willjvsmith (approved):
For reference, the test order we've followed here is:

* Succeeds
* Empty parameter
* Malformed body
* No body
* Empty ID
* Non-existent ID
* String ID

PR #49, opened 2020-03-07
[auth][test] Move request / response types and make tests consistent

jaylees14:
This makes the `auth` service consistent with `user`, as per #47 and #48.

Please compare this with the `user` service directly, we want them to be as close as possible for generation

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -80,7 +80,7 @@ func TestCreateAuthHandlerSucceeds(t *testing.T) {
 		cred,
 	}
 
-	// Make a single auth
+	// Create a single auth
 	res, err := makeRequest(mockEnv, http.MethodPost, "/auth", `{"email": "jay@test.com", "password": "BlackcurrantCrush123"}`)
 	if err != nil {
 		t.Fatalf("Could not make request: %s", err.Error())
@@ -117,7 +117,7 @@ func TestCreateAuthHandlerSucceeds(t *testing.T) {
 	}
 
 	if id.(float64) != 0 {
-		t.Fatalf("ID is incorrect: found %+v, wanted 0", id)
+		t.Fatalf("ID is incorrect, found: %+v, wanted: 0", id)
 	}
 
 	iss, ok := claims["iss"]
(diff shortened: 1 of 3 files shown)


Lewiky (approved):

auth/auth_test.go
```diff
@@ -118,7 +117,7 @@ func TestAuthCreateHandlerSucceeds(t *testing.T) {
 	}
 
 	if id.(float64) != 0 {
-		t.Fatalf("ID is incorrect: found %+v, wanted: 0", id)
+		t.Fatalf("ID is incorrect: found %+v, wanted 0", id)
```
willjvsmith:
```suggestion
		t.Fatalf("ID is incorrect, found: %+v, wanted: 0", id)
```

willjvsmith: and throughout ofc

auth/auth_test.go
```diff
@@ -260,67 +282,86 @@ func TestAuthReadHandlerSucceeds(t *testing.T) {
...
 
-	if res.Code != http.StatusUnauthorized {
+	if res.Code != http.StatusBadRequest {
 		t.Errorf("Wrong status code %v", res.Code)
 	}
 }
 
-// Test that providing an invalid json body causes a `StatusBadRequest`
-func TestAuthReadHandlerFailsOnMalformedJSON(t *testing.T) {
+// Test that providing no body to read endpoint fails
```
willjvsmith:
```suggestion
// Test that providing no body to the read endpoint fails
```

auth/auth_test.go
```diff
@@ -260,67 +282,86 @@ func TestAuthReadHandlerSucceeds(t *testing.T) {
...
-		t.Fatalf("Could not make GET request: %s", err.Error())
+		t.Fatalf("Could not make request: %s", err.Error())
 	}
 
 	if res.Code != http.StatusBadRequest {
+		t.Errorf("Wrong status code: %v", res.Code)
+	}
+}
+
+// Test that providing a non existent auth to the read endpoint fails
```
willjvsmith:
```suggestion
// Test that providing a non-existent auth to the read endpoint fails
```

willjvsmith: and throughout

willjvsmith (approved):
Few small things, also check comments on #48 and #47

PR #50, opened 2020-03-09
[user][auth] Store AuthID alongside user and validate in requests

jaylees14:
We now validate that the same auth that created a "User" can be the only one that can modify or delete it.

We do not explicitly check the auth when reading a user, since we assume that the "User" is a publicly accessible resource to any user with a JWT (@xsanda we should think of a way to express this in Templefile).

I've added unit tests to reflect these changes (yay!)

### DIFF
--- a/user-db/init.sql
+++ b/user-db/init.sql
@@ -1,4 +1,5 @@
 CREATE TABLE User_Temple (
     id SERIAL PRIMARY KEY,
+    auth_id INT NOT NULL,
     name TEXT
 );
--- a/user/user_test.go
+++ b/user/user_test.go
@@ -9,19 +9,25 @@ import (
 	"github.com/TempleEight/spec-golang/user/dao"
 )
 
+// Define 2 JWTs with ID 0 and 1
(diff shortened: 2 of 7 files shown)


c-harding:
# Templefile suggestion

N.b. this means `this` becomes effectively a reserved word. Could alternatively introduce some other syntax, or metadata like `#selfWritable`. Could there also be cases where something can be modified by a certain class of user? E.g. modifiable by all drivers but no passengers?

I don’t think it’s worth the effort of encoding based on user fields, e.g. "visible by a user of experience level at least 6", seems too business-logicky.

image

user-db/init.sql
```diff
@@ -1,4 +1,5 @@
 CREATE TABLE User_Temple (
     id SERIAL PRIMARY KEY,
+    auth_id INT NOT NULL,
```
willjvsmith: absolute anarchy in this file, think I've fixed in #51

willjvsmith (approved):
LGTM 🚀

Lewiky (approved):
LGTM BAYBEE GURL

user/user_test.go
```diff
@@ -9,28 +9,35 @@ import (
 	"github.com/TempleEight/spec-golang/user/dao"
 )
 
+// Define 2 JWTs with ID 0 and 1
+const User0JWT = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1ODMyNTc5NzcsImlkIjowLCJpc3MiOiJmRlM4S21WWXVLQUN5RjN3ZHBQS0hTUXFtWlZWd2pEcSJ9.KzUa-OpHEjFQlsSy7YZI1Kppu4eIU5nyivLvivWcpRc"
```
willjvsmith, line 13: Not UserZeroJWT?

willjvsmith: Missed the large diff on `user_test`

PR #51, opened 2020-03-09
[go-gen] Move request/response types from DAO in Match service, and increase consistency

willjvsmith:
* Makes comments consistent and nice
* Fixes SQL scripts to use camel case table names
* Fixes everything in the Match service to use `int64` for the ID type
* Moves request/response types from DAO in match service
* Fixes DAOs to say `datastore` instead of `database` in comments
* Moves DAO `Init` function to be first function
* Reorders DAO `executeQuery...` functions to be in the same order across services
  * Renames `ExecuteQueryWithResponses`->`ExecuteQueryWithRowResponses`
* Fixes consistent use of `resp` vs service name
* Adds newlines at end of handlers across services to be consistent

What a consistent PR


user/user.go
```diff
@@ -123,7 +124,7 @@ func (env *env) readUserHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	user, err := env.dao.ReadUser(dao.ReadUserInput{
+	match, err := env.dao.ReadUser(dao.ReadUserInput{
```
jaylees14: wat

user/user.go
```diff
@@ -18,29 +18,29 @@ type env struct {
...
 
-// createUserResponse contains the information about the newly created user
+// createUserResponse contains a newly created user to be returned to the client
 type createUserResponse struct {
 	ID   int64
 	Name string
 }
 
-// readUserResponse returns all the information stored about a single user
+// readUserResponse contains a single use to be returned to the client
```
jaylees14:
```suggestion
// readUserResponse contains a single user to be returned to the client
```

match/match_test.go
```diff
@@ -9,112 +9,108 @@ import (
...
-	MatchedOn string
-}
-
 type MockDAO struct {
-	MatchList []MockMatch
+	MatchList []dao.Match
 }
 
 type MockComm struct {
 	UserIDs []int
```
jaylees14:
```suggestion
	UserIDs []int64
```

jaylees14: **consistency**

match/match_test.go
```diff
@@ -9,112 +9,108 @@ import (
...
 	}
-	return dao.ErrMatchNotFound(matchID)
+	return dao.ErrMatchNotFound(input.ID)
 }
 
-func (mc *MockComm) CheckUser(userID int) (bool, error) {
+func (mc *MockComm) CheckUser(userID int64) (bool, error) {
 	for _, id := range mc.UserIDs {
-		if id == userID {
+		if int64(id) == userID {
```
jaylees14: Drop the cast if you apply the change from above

jaylees14 (commented):
LGTM

jaylees14 (approved):

c-harding (approved):
All looks nice

PR #60, opened 2020-03-10
[user][auth] Require email and password for create user endpoint

jaylees14:
First steps in migrating the `auth` service to the `user` service.

This adds the `CREATE` endpoint, which takes a username and password & stores them.

We currently don't return a JWT since the diff would get too large - that'll come next.


jaylees14:
Oh, I forgot to initialise the `authList`. I should do that for consistency. (But it still works, since `append(nil, 5) == []int{5}`)

PR #61, opened 2020-03-11
[api] Add URL parameter comments to the OpenAPI spec

c-harding:
Add a description for the `{id}` in the URL of GET requests


jaylees14 (approved):
LGTM, although may change to `string` soon when we move to `uuid`

Lewiky (approved):

PR #62, opened 2020-03-11
[user][auth] Use ID from JWT when storing user

jaylees14:
- Use the ID provided in the JWT as the primary key for the user table
- Convert ID from int to string in preparation for using UUIDs
- Remove redundant tests about providing a string ID

#### Test Plan
- Existing unit and integration tests updated
- Manual testing:

```
❯❯❯ TOKEN=$(curl -s -X POST localhost:8000/api/auth -d '{"email":"jay@test.com", "password":"BlackcurrantCrush123"}')

❯❯❯ curl -X POST localhost:8000/api/user -d '{"name": "Jay"}' -H "Authorization: Bearer $TOKEN"
{"ID":"1","Name":"Jay"}

❯❯❯ curl -X GET localhost:8000/api/user/1 -H "Authorization: Bearer $TOKEN"
{"ID":"1","Name":"Jay"}
```

PR #63, opened 2020-03-11
[go-gen] Add auth to the Match service

willjvsmith:
* Adds auth checks to endpoints
* Updates tests
  * Changes to populating mock datastore directly rather than with CREATE requests
  * Changes behaviour so that if you attempt to read a match not created by you, it will give an `Unauthorised` error rather than a `Not Found`, i.e. not revealing existence

Should probably add tests for trying to Updating another "creator's" match, but will leave for a later PR (will make a card on backlog).

### DIFF
--- a/match-db/init.sql
+++ b/match-db/init.sql
@@ -1,5 +1,6 @@
 CREATE TABLE match (
     id SERIAL PRIMARY KEY,
+    auth_id INT NOT NULL,
     userOne int,
     userTwo int,
     matchedOn timestamp
--- a/match-db/init.sql
+++ b/match-db/init.sql
@@ -1,7 +1,7 @@
 CREATE TABLE match (
-    id SERIAL PRIMARY KEY,
-    auth_id INT NOT NULL,
-    userOne int,
-    userTwo int,
-    matchedOn timestamp
-  );
+  id SERIAL PRIMARY KEY,
+  auth_id INT NOT NULL,
+  userOne INT,
+  userTwo INT,
+  matchedOn TIMESTAMP,
+);
--- a/match/dao/dao.go
+++ b/match/dao/dao.go
@@ -11,7 +11,7 @@ import (
 
 // Datastore provides the interface adopted by the DAO, allowing for mocking
 type Datastore interface {
-	ListMatch() (*[]Match, error)
+	ListMatch(input ListMatchInput) (*[]Match, error)
 	CreateMatch(input CreateMatchInput) (*Match, error)
 	ReadMatch(input ReadMatchInput) (*Match, error)
 	UpdateMatch(input UpdateMatchInput) (*Match, error)
@@ -30,6 +30,11 @@ type Match struct {
 	UserOne   int64
 	UserTwo   int64
 	MatchedOn string
+}
+
+// ListMatchInput encapsulates the information required to read a match list in the datastore
+type ListMatchInput struct {
+	AuthID int64
 }
 
 // CreateMatchInput encapsulates the information required to create a single match in the datastore
@@ -87,9 +92,9 @@ func executeQueryWithRowResponses(db *sql.DB, query string, args ...interface{})
 	return db.Query(query, args...)
 }
 
-// ListMatch returns a list containing every match in the datastore
-func (dao *DAO) ListMatch() (*[]Match, error) {
-	rows, err := executeQueryWithRowResponses(dao.DB, "SELECT * FROM Match")
+// ListMatch returns a list containing every match in the datastore for a given ID
+func (dao *DAO) ListMatch(input ListMatchInput) (*[]Match, error) {
+	rows, err := executeQueryWithRowResponses(dao.DB, "SELECT * FROM Match WHERE id = $1", input.AuthID)
 	if err != nil {
 		return nil, err
 	}
--- a/match/dao/dao.go
+++ b/match/dao/dao.go
@@ -119,7 +119,7 @@ func (dao *DAO) ListMatch(input ListMatchInput) (*[]Match, error) {
 
 // CreateMatch creates a new match in the datastore, returning the newly created match
 func (dao *DAO) CreateMatch(input CreateMatchInput) (*Match, error) {
-	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO match (auth_id, userOne, userTwo, matchedOn) VALUES ($1, $2, NOW()) RETURNING *", input.UserOne, input.UserTwo)
+	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO match (auth_id, userOne, userTwo, matchedOn) VALUES ($1, $2, $3, NOW()) RETURNING *", input.AuthID, input.UserOne, input.UserTwo)
 
 	var match Match
 	err := row.Scan(&match.ID, &match.AuthID, &match.UserOne, &match.UserTwo, &match.MatchedOn)
--- a/match/dao/dao.go
+++ b/match/dao/dao.go
@@ -26,6 +26,7 @@ type DAO struct {
 // Match encapsulates the object stored in the datastore
 type Match struct {
 	ID        int64
+	AuthID    int64
 	UserOne   int64
 	UserTwo   int64
 	MatchedOn string
@@ -33,6 +34,7 @@ type Match struct {
 
 // CreateMatchInput encapsulates the information required to create a single match in the datastore
 type CreateMatchInput struct {
+	AuthID  int64
 	UserOne int64
 	UserTwo int64
 }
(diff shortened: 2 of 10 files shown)


match-db/init.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE match (
     id SERIAL PRIMARY KEY,
+    auth_id INT NOT NULL,
     userOne int,
     userTwo int,
```
jaylees14: Can we be consistent with the types? `INT` or `int`?

jaylees14: And I'm thinking `created_by` might be better than `auth_id`?

match/match.go
```diff
@@ -105,8 +105,32 @@ func jsonMiddleware(next http.Handler) http.Handler {
...
+	if err != nil {
+		return false, err
+	}
+
+	if match.ID != auth.ID {
+		return false, nil
+	}
+
+	return true, nil
+}
```
jaylees14: Could change this to 

```
return match.ID == auth.ID, nil
```

jaylees14 (commented):
A few comments to start - but probably worth waiting for #62 to go through since the authID goes from an `integer` to `string`. 

W.r.t how that looks in the Templefile - I'm not sure why you'd need `UserOne` to be the same as `authID`? Is there a use case you had in mind?

match-db/init.sql
```diff
@@ -1,6 +1,7 @@
...
-    id SERIAL PRIMARY KEY,
-    userOne int,
-    userTwo int,
-    matchedOn timestamp
-  );
+  id SERIAL PRIMARY KEY,
+  auth_id INT NOT NULL,
+  userOne INT,
+  userTwo INT,
+  matchedOn TIMESTAMP,
```
jaylees14, line 6: Think we've been pretty consistent with the 4 space indentation in these files

match/dao/dao.go
```diff
@@ -112,10 +119,10 @@ func (dao *DAO) ListMatch() (*[]Match, error) {
 
 // CreateMatch creates a new match in the datastore, returning the newly created match
 func (dao *DAO) CreateMatch(input CreateMatchInput) (*Match, error) {
-	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO match (userOne, userTwo, matchedOn) VALUES ($1, $2, NOW()) RETURNING *", input.UserOne, input.UserTwo)
+	row := executeQueryWithRowResponse(dao.DB, "INSERT INTO match (auth_id, userOne, userTwo, matchedOn) VALUES ($1, $2, NOW()) RETURNING *", input.UserOne, input.UserTwo)
```
jaylees14: You're not actually inserting anything new

match/match_test.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/TempleEight/spec-golang/match/dao"
 )
 
+// Define 2 JWTs with ID 0 and 1
+const auth0JWT = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1ODMyNTc5NzcsImlkIjowLCJpc3MiOiJmRlM4S21WWXVLQUN5RjN3ZHBQS0hTUXFtWlZWd2pEcSJ9.KzUa-OpHEjFQlsSy7YZI1Kppu4eIU5nyivLvivWcpRc"
+const auth1JWT = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1ODMyNTc5NzcsImlkIjoxLCJpc3MiOiJmRlM4S21WWXVLQUN5RjN3ZHBQS0hTUXFtWlZWd2pEcSJ9.kXaTT0Yl3-zeWreKOl5Zd6dG1gJG49JSS0zfdBRG_oU"
```
jaylees14, line 14: Let's make a card to update the names of these elsewhere too

match/match_test.go
```diff
@@ -17,30 +21,35 @@ type mockComm struct {
...
+	mockMatchList := make([]dao.Match, 0)
 	for _, match := range md.matchList {
-		matchList = append(matchList, dao.Match{
-			ID:        match.ID,
-			UserOne:   match.UserOne,
-			UserTwo:   match.UserTwo,
-			MatchedOn: match.MatchedOn,
-		})
+		if match.AuthID == input.AuthID {
+			mockMatchList = append(mockMatchList, dao.Match{
```
jaylees14: Confused why you're making a new object here? Why aren't you just appending the `match`?

match/match_test.go
```diff
@@ -97,47 +108,61 @@ func (mc *mockComm) CheckUser(userID int64) (bool, error) {
 	return false, nil
 }
 
-func makeRequest(env env, method string, url string, body string) (*httptest.ResponseRecorder, error) {
+func makeRequest(env env, method string, url string, body string, authToken string) (*httptest.ResponseRecorder, error) {
 	rec := httptest.NewRecorder()
 	req, err := http.NewRequest(method, url, strings.NewReader(body))
+	req.Header.Set("Authorization", "Bearer "+authToken)
```
jaylees14: This should go after checking the error

match/match_test.go
```diff
@@ -97,47 +108,61 @@ func (mc *mockComm) CheckUser(userID int64) (bool, error) {
...
 }
 
-// Test that a match list can be read successfully
+// Test that a match list can be read successfully for a given ID
 func TestListMatchHandlerSucceeds(t *testing.T) {
-	mockEnv := env{
-		&mockDAO{matchList: make([]dao.Match, 0)},
-		&mockComm{userIDs: []int64{0, 1, 2, 3}},
+	// Populate mock datastore
+	matchList := []dao.Match{
```
jaylees14, line 119: Could you modify the list after making the mockEnv? Should mean we can one-line the creation of the empty env

match/match_test.go
```diff
@@ -271,21 +296,23 @@ func TestCreateMatchHandlerFailsOnAllInvalidReferences(t *testing.T) {
...
 func TestReadMatchHandlerSucceeds(t *testing.T) {
+	// Populate mock datastore
+	matchList := []dao.Match{dao.Match{
+		ID:        0,
+		AuthID:    0,
+		UserOne:   0,
+		UserTwo:   1,
+		MatchedOn: "2020-01-01T12:00:00.000000Z",
+	}}
+
```
jaylees14, line 301: Ditto here

jaylees14 (commented):
Nice stuff, just a few consistency issues

match/dao/dao.go
Lewiky: Just a thought here, since we're no longer distinguishing between user_ids and auth_ids, is there ever a case where userOne isn't going to be the same as auth_id?

match-db/init.sql
willjvsmith: Fair, only did it because the Postgres creation scripts we generate indent with 2 spaces

Lewiky (approved):
This was a long boi

match/match_test.go
willjvsmith: copied from user, will change both

match/match_test.go
willjvsmith: Yeah I thought the same, but cannot access the list - discussed offline

match/match_test.go
jaylees14: See #70

match/dao/dao.go
willjvsmith: This is exactly the issue I flagged the other day, can see updates in the cancelled card on Monday.

In brief, there's two ways of modelling the Match service and we should support both.

This implementation:
```
Match: service {
  UserOne: User;
  UserTwo: User;
  ...
}
```

Alternative implementation:
```
Match: service {
  OtherUser: User;
  ...
}
```

`auth_id` will become `created_by` in a future PR.

I'm happy to talk about this more offline, it's quite nuanced and I've got some good examples to illustrate the reason we're doing it this way.

jaylees14: I just had a brief chat with @Lewiky and I think all is clear now

willjvsmith: For future reference:

```
User: service {
  firstName: string;
  lastName: string;
  #auth{by: email};
}
GuineaPig: service {
  name: String;
}
Pairing: service {
  // Implicit auth_id, soon to be changed to created_by
  GuineaPigOne: GuineaPig;
  GuineaPigTwo: GuineaPig;
}
```
here the created_by is always different to the guineaPigOne or Two ID

user/user_test.go
```diff
@@ -71,11 +71,10 @@ func (md *mockDAO) DeleteUser(input dao.DeleteUserInput) error {
 func makeRequest(env env, method string, url string, body string, authToken string) (*httptest.ResponseRecorder, error) {
 	rec := httptest.NewRecorder()
 	req, err := http.NewRequest(method, url, strings.NewReader(body))
-	req.Header.Set("Authorization", "Bearer "+authToken)
 	if err != nil {
 		return nil, err
 	}
-
+	req.Header.Set("Authorization", "Bearer "+authToken)
```
jaylees14, line 70: Bit out of scope for this PR, please try and avoid that in the future.

jaylees14 (approved):

PR #65, opened 2020-03-11
[user][auth] Use google UUID package

jaylees14:
Turns out, there's an official Google package for UUIDs that was **super** hard to find (https://lmgtfy.com/?q=golang+uuid) - so I've switched to that one - the API was basically identical which is nice


Lewiky (approved):

willjvsmith (approved):
LGTM 🚀

PR #66, opened 2020-03-11
[auth] Use UUID for auth service

### LINKED ISSUE #64: this will need to be changed to UUID 0 when our Auth approach is updated
this will need to be changed to UUID 0 when our Auth approach is updated

_Originally posted by @smithwjv in https://github.com/TempleEight/spec-golang/pull/62_

jaylees14:
- Uses `UUID` instead of `int` for the `auth` service
- Closes #64


c-harding (approved):

Lewiky (approved):

willjvsmith (approved):
LGTM 🚀

PR #67, opened 2020-03-12
[auth] Add integration tests for auth

jaylees14:
Add integration tests for the auth service. This required a couple of changes to existing code:

- Rename `utils` package to `util`
- Update the travis script to run the auth integration tests
  - NB: I had to swap the `--abort-on-container-exit` with the name of the service, then use dependencies within the `docker-compose.yml` file to enforce which service we want the output of. The current system became a problem since the `kong-migrations` container exited before the tests would run, causing the whole system to shut down.

NB: the tests will need to be slightly updated for #66


c-harding (approved):
<span title="It all looks very good to me">ialvgtm</abbr>

Lewiky (approved):
LGTM

PR #70, opened 2020-03-12
[user][test] Reorder header setting

jaylees14:
When reviewing #63 I saw this error - assumed it must have come from somewhere else... So I fixed the source 😉


Lewiky (approved):

jaylees14:
Closing because @smithwjv went out of scope

PR #71, opened 2020-03-12
[user] Extract UUID to constant and rename JWTs

jaylees14:
Tidying up existing code to avoid mistakes in tokens

### DIFF
--- a/user/user_it_test.go
+++ b/user/user_it_test.go
@@ -3,6 +3,7 @@
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


c-harding (approved):

willjvsmith (approved):
LGTM 🚀 

Ideally we'd change the tests so they directly add to the mock datastore rather making CREATE requests, but not worth the time to change in regards to marks.

PR #72, opened 2020-03-12
[go-gen] Change IDs to UUIDs and update tests in Match service

willjvsmith:
* Switches out `int64` for `uuid.UUID`
* Updates tests


jaylees14 (approved):
LGTM, although will need to update `user` and `auth` tests to not perform multiple calls in one test as well

c-harding (approved):

PR #75, opened 2020-03-13
[api] Add missing parameter annotations for update/delete blocks

c-harding:


jaylees14 (approved):
LGTM - @smithwjv probably worth merging in these changes

Lewiky (approved):

willjvsmith:
I discovered this while writing for my upcoming PR, thought we had agreed I'd add this in, so I've already done it!

willjvsmith:
Also should be at the path-level as it's same for all child methods

PR #77, opened 2020-03-13
[openapi-gen] Update openapi spec

willjvsmith:
* Updates version to `0.2.0`
* Adds `required` field where applicable
* Adds in whether an endpoint requires an Auth Bearer header
* Updates response errors
* Changes ID types to strings with `uuid` format
  * Not defined in OpenAPI spec, but tools often support it
* Adds `date-time` format to `MatchedOn` field
  * Follows the RFC3339 spec, just as we do
  * Defined in the OpenAPI spec
* Reorders services to Auth, User, Match
* Reorders paths to be in List then CRUD order

Updates OpenAPI specification with latest API changes. I've run it through two validators and seems ok!


api/spec-golang.openapi.yaml
```diff
@@ -1,255 +1,371 @@
...
-    delete:
+        '400':
+          $ref: '#/components/responses/400BadRequest'
+        '401':
+          $ref: '#/components/responses/401Unauthorized'
+        '500':
+          $ref: '#/components/responses/500InternalServerError'
+  /user/{id}:
+    parameters:
+          - in: path
```
jaylees14: Indentation looks a bit off here?

api/spec-golang.openapi.yaml
```diff
@@ -1,255 +1,371 @@
...
+                    format: date-time
+        '400':
+          $ref: '#/components/responses/400BadRequest'
+        '401':
+          $ref: '#/components/responses/401Unauthorized'
+        '500':
+          $ref: '#/components/responses/500InternalServerError'
+  /match/{id}:
+    parameters:
+        - in: path
```
jaylees14: Ditto here

jaylees14 (approved):
LGTM

Lewiky (approved):

PR #80, opened 2020-03-16
[user][hook] Add hook for create endpoint

jaylees14:
The first prototype of adding additional logic via hooks. I've left the implementation empty, but an example of how it could be used is below.

For now, the **before** calls takes the env (to allow for more logging and metrics later on), a **value** type of the request and **reference** type of the dao input. The request type is not a pointer, since any changes to this object will not be used anywhere, so we can implicitly indicate that through the type - however we still want access to it for any `@client` attributes.

The DAO type can be modified and this will change what's stored in the datastore, so this is a pointer - this can be used for setting `@serverSet` attributes

Please look at the filenames / type names and let me know what you think. Definitely aware these will probably change as we want to add more functionality, but I think it's a good starting point.

Example Hooks:
``` go
package main

import (
	"errors"
	"log"
	"net/http"

	"github.com/TempleEight/spec-golang/user/dao"
)

func (env *env) setup() {
	env.hook.beforeCreate(beforeCreateHook)
	env.hook.afterCreate(afterCreateHook)
}

func beforeCreateHook(env *env, req createUserRequest, input *dao.CreateUserInput) *HookError {
	if req.Name == "Lewis" {
		return &HookError{http.StatusTeapot, errors.New("we do not allow users called Lewis")}
	}
        // All users are now named Corona
	input.Name = "Corona"
	return nil
}

func afterCreateHook(env *env, user *dao.User) *HookError {
	log.Printf("Added a user with ID: %s", user.ID)
	return nil
}
```

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -17,6 +17,7 @@ import (
 // env defines the environment that requests should be executed within
 type env struct {
 	dao dao.Datastore
+	hook Hook
 }
 
 // createUserRequest contains the client-provided information required to create a single user
@@ -74,7 +75,10 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	env := env{d}
+	env := env{d, Hook{}}
+
+	// Call into non-generated entry-point
+	env.setup()
 
 	log.Fatal(http.ListenAndServe(":80", env.router()))
 }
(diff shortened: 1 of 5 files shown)


willjvsmith (approved):
LGTM 🚀

This is pretty cool, adds really interesting functionality - will be great to write about!

Lewiky (approved):
I like it! Very clean and elegant

user/user.go
```diff
@@ -110,16 +114,36 @@ func (env *env) createUserHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	user, err := env.dao.CreateUser(dao.CreateUserInput{
+	input := dao.CreateUserInput{
 		ID:   auth.ID,
 		Name: req.Name,
-	})
+	}
+
+	for _, hook := range env.hook.beforeCreateHooks{
```
c-harding: Is this really the gofmt-approved spacing before the open-brace?

user/user_test.go
```diff
@@ -81,6 +81,7 @@ func makeRequest(env env, method string, url string, body string, authToken stri
 func makeMockEnv() env {
 	return env{
 		&mockDAO{userList: make([]dao.User, 0)},
```
c-harding, line 83: Didn’t we extract this `make` stuff away recently?

c-harding (approved):
Looks very nice! Couple of comments

user/user.go
jaylees14: Good catch, IntelliJ doesn't have gofmt turn on for save </3

user/user_test.go
jaylees14: We did... into this function ;)

PR #82, opened 2020-03-16
[user][hooks] Allow router configuration

jaylees14:
By taking the router as an argument to set up, the user can define their own additional endpoints:

```go
func (env *env) setup(router *mux.Router) {
    router.HandleFunc("/yeet", env.yeetHandler).Methods(http.MethodGet)
}
```

Unfortunately the router doesn't let you delete existing endpoints, so to accomplish this you'd have to create a new router and add only the endpoints you want using my *favourite* `go` syntax...

```go
func (env *env) setup(router *mux.Router) {
    *router = *mux.NewRouter()
    router.HandleFunc("/yeet", env.readUserHandler).Methods(http.MethodGet)
}
```

### DIFF
--- a/user/setup.go
+++ b/user/setup.go
@@ -1,5 +1,7 @@
 package main
 
-func (env *env) setup() {
+import "github.com/gorilla/mux"
+
+func (env *env) setup(router *mux.Router) {
 	// Add user defined code here
 }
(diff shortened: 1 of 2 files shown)


willjvsmith (approved):
horrible

Lewiky (approved):

PR #83, opened 2020-03-16
[hooks] Extend datastore type

jaylees14:
- Add a new DAO type in every service that extends the generated interface
- Fix the CI, since it would exit with the status code of the last loop iteration, instead of early returning


c-harding (approved):

willjvsmith (approved):
LGTM 🚀 

Probs should rename PR name without [user]

PR #89, opened 2020-03-18
[metrics] Add bare bones metrics config, with grafana and prometheus

jaylees14:
Adds a very simple Grafana and Prometheus config, with a single dashboard.

If you spin up the `docker-compose` environment, Grafana is exposed on `3000`, default username and password is `admin`.

Don't worry too much about the `.json` file, this will change as we add more metrics - the `.yml` files should remain the same though.


Lewiky (approved):
Nice one

user/go.sum
```diff
@@ -1,10 +1,96 @@
...
+github.com/go-logfmt/logfmt v0.4.0/go.mod h1:3RMwSq7FuexP4Kalkev3ejPJsZTpXXBr9+V4qmtdjCk=
+github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
+github.com/gogo/protobuf v1.1.1/go.mod h1:r8qH/GZQm5c6nD/R0oafs1akxWv10x8SbQlK7atdtwQ=
+github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/golang/protobuf v1.3.1/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/golang/protobuf v1.3.2 h1:6nsPYzhq5kReh6QImI3k5qWzO4PEbvbIW2cwSfR/6xs=
+github.com/golang/protobuf v1.3.2/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
+github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/gofuzz v1.0.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
```
willjvsmith, line 29: nice

user/user.go
```diff
@@ -60,6 +61,12 @@ func defaultRouter(env *env) *mux.Router {
 }
 
 func main() {
+	// Prometheus metrics
+	go func() {
+		http.Handle("/metrics", promhttp.Handler())
+		http.ListenAndServe(":2112", nil)
```
willjvsmith, line 67: Prometheus always on 2112? If not do we wanna make it a config field?

willjvsmith (approved):
LGTM 🚀

user/user.go
jaylees14: Good shout, I guess we should do this for the service port too

jaylees14: Will add a card

PR #90, opened 2020-03-18
[metrics][user] Add custom metrics to the user service

jaylees14:
- Add some custom metrics to the user service
- Show success and failure rate charts, averaged over the last 5 mins

The way in which you add a failure metric is kinda annoying, there's lots of repeated code - I'm gonna pull this out into a utility function before adding any more.

Also apologies for the JSON.. looks like I've just found what I'm generating for the next few days ❤️ 

Currently looks like:

Screenshot 2020-03-18 at 16 20 06

### DIFF
--- a/user/metrics.go
+++ b/user/metrics.go
@@ -1,0 +1,23 @@
+package main
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	requestCreate = "create"
+	requestRead   = "read"
+	requestUpdate = "update"
+	requestDelete = "delete"
+
+	requestSuccess = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "user_request_success_total",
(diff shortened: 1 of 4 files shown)


user/metrics.go
```diff
@@ -0,0 +1,23 @@
+package main
```
jaylees14: Not sure if this should go in a separate package @smithwjv?

willjvsmith: Yeah stick it in a new one I reckon

willjvsmith (approved):
LGTM 🚀 

Surprised no other logs were recommended. Perhaps database accesses? Or when u get round to Match, interservice-comms would be a good one

jaylees14:
> LGTM 🚀
> 
> Surprised no other logs were recommended. Perhaps database accesses? Or when u get round to Match, interservice-comms would be a good one

Oh this isn't by all means all of them - just a couple of easy ones to get us started 👍

c-harding (approved):

PR #91, opened 2020-03-18
[user] Replace error responses with util function

jaylees14:
Turns out, writing the same thing over and over again is **really** annoying.
I think this is what they call _good software development_...

I also found that a `http.Error` sets the response header to Plain Text (https://golang.org/pkg/net/http/#Error), which ruins our `jsonMiddleware`, so I've replaced it with only the core functionality we want from the `http.Error` function, see: 

https://github.com/golang/go/blob/f1f947af28d1de655fe7eb845a053753d10845aa/src/net/http/server.go#L2021

#NegativeDiffSqad

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -62,6 +62,13 @@ func defaultRouter(env *env) *mux.Router {
 	return r
 }
 
+// respondWithError responds to a HTTP request with a JSON error response
+func respondWithError(w http.ResponseWriter, err string, statusCode int, requestType string) {
+	w.WriteHeader(statusCode)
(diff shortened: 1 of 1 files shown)


c-harding (approved):
nice

Lewiky (approved):
respondWithError(w, Error("LGTM!"), 200, metric.RequestDelete)

willjvsmith (approved):
Nice, good improvement 🚀 

Could move `respondWithError` to `util.go` but I guess the only place we actually respond to the request is in this file, so no strong feelings

#NegativeDiffSqad

PR #92, opened 2020-03-19
[metrics][user] Add database query time metric

jaylees14:
- Add a new metric to measure DB request times, bucketed into 50th, 90th, 95th and 99th percentiles. 

Screenshot 2020-03-19 at 10 23 36


Lewiky (approved):
YAH

c-harding (approved):

PR #93, opened 2020-03-19
[match] Add error util function

jaylees14:
Same as #91, but for match

#NegativeDiffSqad

### DIFF
--- a/match/match.go
+++ b/match/match.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/TempleEight/spec-golang/match/comm"
 	"github.com/TempleEight/spec-golang/match/dao"
+	"github.com/TempleEight/spec-golang/match/metric"
 	"github.com/TempleEight/spec-golang/match/util"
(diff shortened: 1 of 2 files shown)


match/match.go
```diff
@@ -77,6 +78,12 @@ func defaultRouter(env *env) *mux.Router {
 	return r
 }
 
+// respondWithError responds to a HTTP request with a JSON error response
+func respondWithError(w http.ResponseWriter, err string, statusCode int, requestType string) {
+	w.WriteHeader(statusCode)
+	fmt.Fprintln(w, util.CreateErrorJSON(err))
```
Lewiky, line 84: How come you're not adding the metric for failure here like in the other service?

jaylees14: That was just so the PR size wouldn't balloon, since none of the metrics stuff is set up for the match service yet. *Coming soon* 😉

Lewiky (approved):
LGTM bruh

c-harding (approved):

PR #95, opened 2020-03-19
[metrics][match] Add metrics to match service

jaylees14:
Add some metrics to the ol' match service. This should bring it up to the same level as the user service - they're essentially identical except for the addition of the `read` endpoint 😄


c-harding (approved):
IMG_3558

Lewiky (approved):
Assuming the JSON is right LGTM

PR #97, opened 2020-03-19
[Infra] Add travis stage to build and push images

Lewiky:
This adds a script and associated travis build stage for building docker images and pushing them to the registry on push to master. 

Also the shell script will be generated as a part of the kube-gen step

## Test plan:

Run `sh push-image.sh` after setting env vars and verify that the images are pushed correctly to the registry.


.travis.yml
```diff
@@ -2,8 +2,13 @@ sudo: required
...
-# Run unit and integration tests
-- cd user && docker-compose up --build user && cd ..
-- cd auth && docker-compose up --build auth && cd ..
-- cd match && docker-compose up --build match && cd ..
+jobs:
+  include:
+  - stage: test
+    name: tests
+    # Run unit and integration tests
+    script: for svc in user auth match; do cd $svc && docker-compose up --build $svc && cd ..; done
```
jaylees14: This can’t be a for loop - we need it to fail if any of the iterations fail and a loop only returns the exit code of the last iteration

push-image.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+if [[ -z "${DOCKER_USERNAME}" ]] || [[ -z "${DOCKER_PASSWORD}" ]]; then
+  echo "Please set DOCKER_USERNAME and DOCKER_PASSWORD variables"
+  exit 1
+fi
+
+
+for service in "user" "auth" "match"; do
+  docker build -t "registry.lewiky.com/temple-$service-service" $service
```
jaylees14: Could we pull the base url into a variable so it’s easy to swap if needed?

.travis.yml
Lewiky: [Travis only allows one command per script block](https://docs.travis-ci.com/user/deployment/script/) (at least in stages), so I've separated them out into different build stages - also means they can be parallelised

jaylees14 (approved):
LGTM 👍

push-image.sh
```diff
@@ -0,0 +1,13 @@
...
+  echo "Please set DOCKER_USERNAME and DOCKER_PASSWORD variables"
+  exit 1
+fi
+
+REGISTRY_URL="registry.lewiky.com"
+
+for service in "user" "auth" "match"; do
+  docker build -t "$REGISTRY_URL/temple-$service-service" $service
+  docker push "$REGISTRY_URL/temple-$service-service"
+done
```
c-harding, line 13: newline at the end of the file plz

c-harding (approved):

PR #98, opened 2020-03-19
[metrics][auth] Add metrics for auth service

jaylees14:
- Equivalent metrics for the auth service, copying the `user` and `match` service


Lewiky (approved):
Cool man

c-harding (approved):
“Failure” is in red

PR #99, opened 2020-03-24
[config] Move ports to config file

jaylees14:
Move ports to config file, rather than hard coding them in the source code

### DIFF
--- a/match/match.go
+++ b/match/match.go
@@ -89,12 +89,6 @@ func respondWithError(w http.ResponseWriter, err string, statusCode int, request
 }
 
 func main() {
-	// Prometheus metrics
-	go func() {
-		http.Handle("/metrics", promhttp.Handler())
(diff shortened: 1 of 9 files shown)


willjvsmith (approved):
LGMT 🚀 

We're gonna need to put some thought into choosing ports in the mapping

match/match.go
```diff
@@ -106,6 +100,16 @@ func main() {
 		log.Fatal(err)
 	}
 
+	// Prometheus metrics
+	promPort, ok := config.Ports["prometheus"]
+	if !ok {
+		log.Fatal(err)
```
c-harding: Do we want to give a more specific error, or is this message understandable enough?

c-harding (approved):

match/match.go
jaylees14: Ha! The err here isn't even valid - good spot

PR #101, opened 2020-03-31
[user] Add picture struct create endpoint

jaylees14:
Add support for a `picture` struct in the `user` service, which involves a couple of design decisions:

- Adds a new endpoint `user/<id>/picture` with CRUD defined on that (C now, RUD coming later)
- Implicit foreign key constraint on the picture table, but never returned
- Whenever a byte array is used it is NOT returned in the response body
- Byte arrays are expected to be encoded into base64 whenever used in JSON

Have added unit tests for this endpoint, but for those that like manual testing too:

```bash
ACCESS1=$(curl -s -X POST $BASE/api/auth/register -d "{\"email\":\"jay$RANDOM@test.com\", \"password\":\"abcdefgh\"}" | jq -r .AccessToken)

USER1=$(curl -s -X POST $BASE/api/user -d '{"name": "Jay"}' -H "Authorization: Bearer $ACCESS1" | jq -r .ID)

CAT=$(base64 ~/cat.png)

echo "{ \"img\": \"$CAT\" }" | curl -X POST $BASE/api/user/$USER1/picture -H "Authorization: Bearer $ACCESS1" -d @-

```

### DIFF
--- a/user/dao/dao.go
+++ b/user/dao/dao.go
@@ -3,6 +3,7 @@ package dao
 import (
 	"database/sql"
 	"fmt"
+	"github.com/lib/pq"
 
 	"github.com/TempleEight/spec-golang/user/util"
 	// pq acts as the driver for SQL requests
@@ -10,12 +11,16 @@ import (
 	_ "github.com/lib/pq"
 )
 
+// https://www.postgresql.org/docs/9.3/errcodes-appendix.html
+const psqlForeignKeyViolation = "foreign_key_violation"
+
 // BaseDatastore provides the basic datastore methods
 type BaseDatastore interface {
 	CreateUser(input CreateUserInput) (*User, error)
 	ReadUser(input ReadUserInput) (*User, error)
 	UpdateUser(input UpdateUserInput) (*User, error)
 	DeleteUser(input DeleteUserInput) error
+	CreatePicture(input CreatePictureInput) (*Picture, error)
 }
 
 // DAO encapsulates access to the datastore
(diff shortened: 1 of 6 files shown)


Lewiky (approved):
Cool cool cool

user/dao/dao.go
```diff
@@ -136,3 +155,22 @@ func (dao *DAO) DeleteUser(input DeleteUserInput) error {
 
 	return nil
 }
+
+// CreatePicture new picture in the datastore, returning the newly created picture
```
willjvsmith, line 159: Do we want the DAO to return the picture?

user/user.go
```diff
@@ -358,3 +370,89 @@ func (env *env) deleteUserHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	userID, err := util.ExtractIDFromRequest(mux.Vars(r))
+	if err != nil {
+		respondWithError(w, err.Error(), http.StatusBadRequest, metric.RequestCreatePicture)
+		return
+	}
+
+	// Only the auth that created the user can create a single picture
```
willjvsmith, line 387: Strange sentence, but I understand why 😄

willjvsmith (approved):
LGTM 🚀 

Would have liked a `cat.png` for the manual test

user/dao/dao.go
jaylees14: I just did it for consistency, but agree we probably don't need it. Can remove it - what do you think?

PR #103, opened 2020-04-01
[kube-metrics] Add metrics to kube

Lewiky:
Who loves reviewing yaml!!! I do !!!

Adds prometheus and grafana to el kubernooties

## Test plan
* `source deploy.sh`

* `minikube service grafana`

* make some requests as usual and watch them come in the grafana dash

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -32,8 +32,8 @@ kubectl create configmap grafana-datasource-config --from-file "$BASEDIR/grafana
 
 for file in "$BASEDIR/grafana/provisioning/dashboards/"*
 do
-  filename="${file##*/}" # Get everything after the final /
-  file="${filename%.*}" # Get everything before the . (filename without extension)
+  filename=$(basename $file) # Get everything after the final /
+  file=$(echo $filename | cut -f 1 -d '.') # Get everything before the . (filename without extension)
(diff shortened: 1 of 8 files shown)


deploy.sh
```diff
@@ -14,10 +14,29 @@ minikube start --vm-driver=virtualbox
...
 
+# Prometheus
+kubectl create configmap prometheus-file-config --from-file "$BASEDIR/prometheus/prometheus.yml" -o=yaml
+
+# Grafana
+kubectl create configmap grafana-datasource-config --from-file "$BASEDIR/grafana/provisioning/datasources/datasource.yml" -o=yaml
+
+for file in "$BASEDIR/grafana/provisioning/dashboards/"*
+do
+  filename="${file##*/}" # Get everything after the final /
```
jaylees14: Using bash substrings, you brave boy 😉  I think you might need `/bin/bash` at the top for this always to be correct.

jaylees14: Also.. what happens if there's a folder in this directory?

jaylees14 (commented):
Changes work locally for me, just a small comment in the shell script

c-harding (approved):
There‘s witchcraft, but it’s documented, so sure

jaylees14 (approved):
LGTM

PR #105, opened 2020-04-03
[auth] Rename hooks to register and login

jaylees14:
To be in keeping with the endpoint names in the auth service, rename the hooks to `register` and `login` rather than `create` and `read`

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -200,12 +200,12 @@ func TestCreateAuthHandlerFailsOnDuplicate(t *testing.T) {
 	}
 }
 
-// Test that a before create hook is successfully invoked
+// Test that a before register hook is successfully invoked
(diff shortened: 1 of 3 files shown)


Lewiky (approved):

auth/auth_test.go
```diff
@@ -461,12 +461,12 @@ func TestReadAuthHandlerBeforeHookAbortsRequest(t *testing.T) {
 	}
 }
 
-// Test that an after create hook is successfully invoked
+// Test that an after login hook is successfully invoked
```
willjvsmith, line 464: Was this wrong before?

willjvsmith (approved):
LGTM 🚀

auth/auth_test.go
jaylees14: Hahah yep

PR #108, opened 2020-04-05
[match] Remove server set values from SQL query and use hooks instead

jaylees14:
Rather than using `NOW()` in Postgres, we've decided to pull all `@serverSet` values into the Hooks. This means that without providing a hook, the value stored will be the Go default.

These are usually pretty sensible (0, "", 0.0f etc.), with `time.Time` defaulting to *January 1st 0001 at 00:00.00*. I think this makes it clear to the end developer that they need to update this value somewhere, rather than us defaulting to `time.Now()`, even if that is probably the most common use case. 

This also gives a nice test of hooks, which from a bit of manual testing seem to work nicely:

```
BASE=${KONG_ENTRY:-"localhost:8000"}

ACCESS1=$(curl -s -X POST $BASE/api/auth/register -d "{\"email\":\"jay$RANDOM@test.com\", \"password\":\"abcdefgh\"}" | jq -r .AccessToken)
ACCESS2=$(curl -s -X POST $BASE/api/auth/register -d "{\"email\":\"jay$RANDOM@test.com\", \"password\":\"abcdefgh\"}" | jq -r .AccessToken)

USER1=$(curl -s -X POST $BASE/api/user -d '{"name": "Jay"}' -H "Authorization: Bearer $ACCESS1" | jq -r .ID)
USER2=$(curl -s -X POST $BASE/api/user -d '{"name": "Lewis"}' -H "Authorization: Bearer $ACCESS2" | jq -r .ID)

MATCH=$(curl -X POST $BASE/api/match -d "{\"UserOne\": \"$USER1\", \"UserTwo\": \"$USER2\"}" -H "Authorization: Bearer $ACCESS1" | jq -r .ID)
curl -s -X GET $BASE/api/match/$MATCH -H "Authorization: Bearer $ACCESS1" 
```
Without the hook produces:
```
{"ID":"...","UserOne":"...","UserTwo":"...","MatchedOn":"0001-01-01T00:00:00Z"}
```

With the hook produces (noting this time is in UTC):
```
{"ID":"...","UserOne":"...","UserTwo":"...","MatchedOn":"2020-04-05T11:50:35Z"}
```

### DIFF
--- a/match/dao/dao.go
+++ b/match/dao/dao.go
@@ -42,10 +42,11 @@ type ListMatchInput struct {
 
 // CreateMatchInput encapsulates the information required to create a single match in the datastore
 type CreateMatchInput struct {
-	ID      uuid.UUID
-	AuthID  uuid.UUID
-	UserOne uuid.UUID
-	UserTwo uuid.UUID
+	ID        uuid.UUID
+	AuthID    uuid.UUID
+	UserOne   uuid.UUID
+	UserTwo   uuid.UUID
+	MatchedOn time.Time
 }
 
 // ReadMatchInput encapsulates the information required to read a single match in the datastore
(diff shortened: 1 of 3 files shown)


willjvsmith (approved):
LGTM 🚀 this project really coming together

Manual testing worked for me 👍

c-harding (approved):

